PR #6, opened 2022-08-02
Wasm calls

RolandMacDoland:
This PR addresses the addition of WASM functionality both in the parser and the generator to handle WASM calls back and forth WASM ops.


.github/workflows/test.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     strategy:
       matrix:
         python-version: ['3.8', '3.9', '3.10']
-        os: ['ubuntu-20.04', 'macos-11', 'windows-2019']
+        os: ['macos-11']
```
cqc-alec: Why remove these?

.github/workflows/test.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     strategy:
```
cqc-alec: Please remove all the wheel files, these should not be in version control.

pytket_qir/qir.py
```diff
@@ -12,16 +12,22 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""This module contains all functionality to parse and generate QIR files."""
+"""
+This module contains all functionality to parse and generate QIR files
+from and to pytket circuits.
```
cqc-alec:
```suggestion
to and from pytket circuits.
```

pytket_qir/qir.py
```diff
@@ -228,6 +240,25 @@ def block_to_circuit(self, block: QirBlock, circuit: Circuit) -> Circuit:
                 circuit.add_gate(op, unitids)
             elif instr.instr.is_qir_call:  # Setting the conditional bit for branching.
                 c_reg_index = instr.instr.call_func_params[0].constant.result_static_id
+            elif instr.instr.is_call:  # WASM external call.
+                if not self.wasm_handler:
+                    raise ValueError("A WASM file handler must be provided.")
+                func_name = instr.func_name
+                assert func_name
+                matched_str = re.search("__quantum__(.+?)__(.+?)__(.+)", func_name)
+                assert matched_str
```
cqc-alec: I think these asserts should be exceptions, as the conditions are outside the control of this module?

pytket_qir/qir.py
```diff
@@ -400,11 +445,80 @@ def condition_zero_block():
...
+            inputs: List[str] = []
+            outputs: List[str] = []
+            for reglist, sizes in [
+                (inputs, op.input_widths),
+                (outputs, op.output_widths),
+            ]:
+                for in_width in sizes:
+                    wasm_bits = args[:in_width]
+                    args = args[in_width:]
+                    regname = wasm_bits[0].reg_name
```
cqc-alec, line 499: Should handle the degenerate case where `in_width` is zero.

pytket_qir/qir.py
```diff
@@ -400,11 +445,80 @@ def condition_zero_block():
...
+                    if wasm_bits != list(cregs[regname]):
+                        WASMUnsupportedError("WASM ops must act on entire registers.")
+                    reglist.append(regname)
+
+            reg = circ.get_c_register(inputs[0])
```
cqc-alec: I don't see where `inputs` has been populated?

pytket_qir/qir.py
```diff
@@ -400,11 +445,80 @@ def condition_zero_block():
...
+                return_type=types.Int(64),
+            )
+
+            # Update gateset in module.
+            module.gateset = PYQIR_GATES
+
+            # A utility function to convert from a pytket
+            #  BitRegister to an SSA variable via pyqir types.
+            reg2var = module.module.add_external_function(
+                "reg2var", types.Function([types.BOOL] * 64, types.Int(64))
```
cqc-alec: Does this only work with registers of size 64? Should we also handle smaller registers by zero-padding?

cqc-alec (changes requested):

.github/workflows/test.yml
RolandMacDoland: Because I used locally compiled wheels for MacOS. This is temporary until the non-void return types and classical ops are released in `pyqir`. Currently, this is merged in the `main` branch only.

.github/workflows/test.yml
RolandMacDoland: Sure. I will do as soon as there is a new `pyqir` release.

pytket_qir/qir.py
RolandMacDoland: Thanks.

pytket_qir/qir.py
RolandMacDoland: OK, got that. More generally, do you mean that if conditions are external to the module for it to possibly recover, then that should be an exception ?

RolandMacDoland: Done.

pytket_qir/qir.py
RolandMacDoland: `reglist, sizes` iterates over the tuples formed by the `(registers, widths)` here:

```
for reglist, sizes in [
                (inputs, op.input_widths),
                (outputs, op.output_widths),
            ]:
``` 

To be honest, this bit of code is brutally nicked from Seyon's code about handling WASM in `pytket.qasm` [here](https://github.com/CQCL/tket/blob/dc2136fb621b3d1659d5e98c73138564dbb0fa39/pytket/pytket/qasm/qasm.py#L1010).

cqc-alec: Ah, it's OK, `reglist` is of course a reference to `inputs` and it does get updated.

pytket_qir/qir.py
RolandMacDoland: Done.

pytket_qir/qir.py
RolandMacDoland: Atm, yes. I can check on the size of `reg` and zero-pad `reg_bool` if necessary.

RolandMacDoland: Done.

pytket_qir/qir.py
```diff
@@ -400,11 +450,82 @@ def condition_zero_block():
...
+            args = command.args
+
+            inputs: List[str] = []
+            outputs: List[str] = []
+            for reglist, sizes in [
+                (inputs, op.input_widths),
+                (outputs, op.output_widths),
+            ]:
+                if not sizes:
+                    ValueError("WASM input or output registers have empty widths.")
```
cqc-alec: Is this an error? Is it invalid WASM?

pytket_qir/qir.py
```diff
@@ -400,11 +450,82 @@ def condition_zero_block():
...
+
+            # Update gateset in module.
+            module.gateset = PYQIR_GATES
+
+            # A utility function to convert from a pytket
+            #  BitRegister to an SSA variable via pyqir types.
+            reg2var = module.module.add_external_function(
+                "reg2var", types.Function([types.BOOL] * 64, types.Int(64))
+            )
+            bool_reg = list(map(bool, bit_reg)) + [False] * (64 - bit_reg.size)
```
cqc-alec: What if `bit_reg.size > 64`?

.github/workflows/test.yml
cqc-alec: I still don't think wheel files should be in source control. If the current code doesn't work with the released version of pyqir, it shouldn't be on the develop branch (or it should build its own version). The idea of the CI is that it should be possible to do a release from develop provided all tests pass -- if that isn't the case then this shouldn't be pushed there.

cqc-alec (commented):
As explained, I think we should wait for a compatible release of pyqir before merging this. I don't think it makes sense to merge something that is dependent on locally-built wheels and that only works on one platform.

RolandMacDoland:
> As explained, I think we should wait for a compatible release of pyqir before merging this. I don't think it makes sense to merge something that is dependent on locally-built wheels and that only works on one platform.

I have no problem with that. I just didn't want to delay this work any longer by waiting for a `pyqir` release that will happen at some point in time which is out of my control. So this can stay in its own branch provided that it has been approved until the release is out and that I can re-instate a proper dependency.

.github/workflows/test.yml
RolandMacDoland: I guess we agreed here.

pytket_qir/qir.py
RolandMacDoland: Hm, not sure what you mean. The idea here is to check whether there isn't any empty input registers. Although it might be permitted to input empty registers, it feels like there should be a check at some point.

pytket_qir/qir.py
RolandMacDoland: I guess there should be some narrowing of the register ? But that should come at the price of information loss ?

RolandMacDoland: Please, bear in mind that I've also changed this a bit in the next PR.

RolandMacDoland: Done.

pytket_qir/qir.py
RolandMacDoland: Done.

bamarsha:
@Roland-djee PyQIR 0.6.0 is released on PyPI now.

pytket_qir/gatesets/base.py
```diff
@@ -1,17 +1,19 @@
...
 from typing import Callable, Dict, List, NamedTuple, Union
 
-from pyqir.generator.types import Double, Integer, Qubit, Result, Void  # type: ignore
+from pyqir.generator.types import Double, Int, Qubit, Result, Void  # type: ignore
 
 
-PyQirInputTypes = Union[Double, Qubit, Result]
-PyQirOutputTypes = Union[Integer, Result, Void]
+PyQirParameterTypes = Union[Double, Qubit, Result]
+PyQirReturnTypes = Union[Int, Result, Void]
```
cqc-alec: The names would be better as singular:
```suggestion
PyQirParameterType = Union[Double, Qubit, Result]
PyQirReturnType = Union[Int, Result, Void]
```

tests/qir_test_files/WASM.ll
```diff
@@ -0,0 +1,14 @@
+source_filename = "Pytket circuit"
```
cqc-alec: Is this right?

pytket_qir/qir.py
```diff
@@ -87,14 +93,22 @@
 }
 
 
+class WASMUnsupportedError(Exception):
```
cqc-alec: Can you document what this means? I assume it doesn't mean WASM is completely unsupported? Maybe `WASMError` would be a better name?

pytket_qir/qir.py
```diff
@@ -228,6 +241,29 @@ def block_to_circuit(self, block: QirBlock, circuit: Circuit) -> Circuit:
                 circuit.add_gate(op, unitids)
             elif instr.instr.is_qir_call:  # Setting the conditional bit for branching.
                 c_reg_index = instr.instr.call_func_params[0].constant.result_static_id
+            elif instr.instr.is_call:  # WASM external call.
+                if not self.wasm_handler:
+                    raise ValueError("A WASM file handler must be provided.")
+                func_name = instr.func_name
+                if not func_name:
```
cqc-alec: What are the values this is intended to catch? Better to be explicit, e.g. `""` or `None`.

pytket_qir/qir.py
```diff
@@ -228,6 +241,29 @@ def block_to_circuit(self, block: QirBlock, circuit: Circuit) -> Circuit:
...
+                    raise ValueError("The WASM function call is not defined.")
+                matched_str = re.search("__quantum__(.+?)__(.+?)__(.+)", func_name)
+                if not matched_str:
```
cqc-alec: Similarly here.

pytket_qir/qir.py
```diff
@@ -228,6 +241,29 @@ def block_to_circuit(self, block: QirBlock, circuit: Circuit) -> Circuit:
...
+                    raise ValueError(
+                        "The WASM function call name is not propely defined."
+                    )
+                # WASM function call parameters.
+                param_regs = []
+                for c_reg_index in range(len(instr.func_args)):
+                    c_reg_name = "c_reg_wasm" + str(c_reg_index)
+                    param_regs.append(circuit.add_c_register(c_reg_name, 64))
```
cqc-alec: Can WASM variables only be 64-bit?

pytket_qir/qir.py
```diff
@@ -300,42 +338,54 @@ def circuit_from_qir(
...
-                self.__setattr__(
-                    v.opname.value,
-                    self.module.add_external_function(
-                        self.gateset.template.substitute(
-                            opnat=v.opnat.value,
-                            opname=v.opname.value,
-                            opspec=v.opspec.value,
-                        ),
-                        types.Function(v.function_signature, v.return_type),
+        self.gateset = gateset if gateset else PYQIR_GATES
```
cqc-alec:
```suggestion
        self.gateset = gateset if gateset is not None else PYQIR_GATES
```

pytket_qir/qir.py
```diff
@@ -300,42 +338,54 @@ def circuit_from_qir(
...
+        self.wasm_handler = wasm_handler
+
+    @property
+    def gateset(self):
+        """A getter for the gateset."""
+        assert self._gateset
```
cqc-alec:
```suggestion
        assert self._gateset is not None
```

pytket_qir/qir.py
```diff
@@ -400,11 +463,76 @@ def condition_zero_block():
...
+            args = command.args
+
+            inputs: List[str] = []
+            outputs: List[str] = []
+            for reglist, sizes in [
+                (inputs, op.input_widths),
+                (outputs, op.output_widths),
+            ]:
+                for in_width in sizes:
+                    assert in_width > 0
```
cqc-alec: Can't we have `in_width == 0` or is that invalid WASM?

pytket_qir/gatesets/base.py
```diff
@@ -45,60 +48,17 @@ class OpSpec(Enum):
...
-        (
-            "return_type",
-            PyQirOutputTypes,
-        ),
-    ],
-)
+@dataclass(frozen=True)
+class QirGate:
+    opnat: OpNat
+    opname: Enum
```
cqc-alec: Why does this have to be an `Enum` rather than just say a string? Given that it can be _any_ enum I'm not sure what the advantage is?

pytket_qir/qir.py
```diff
@@ -457,14 +572,23 @@ def write_qir_file(
...
 def circuit_to_qir_bytes(
-    circ: Circuit, gateset: Optional[CustomGateSet] = None
+    circ: Circuit,
+    gateset: Optional[CustomGateSet] = None,
+    wasm_path: Optional[Union[str, os.PathLike]] = None,
 ) -> bytes:
-    """Return a pytket circuit as bytes."""
+    """Return a pytket circuit as an IR bitcode file."""
+    wasm_handler = None
+    if wasm_path:
```
cqc-alec:
```suggestion
    if wasm_path is not None:
```

tests/conftest.py
```diff
@@ -24,7 +24,7 @@
     OpType,
 )
 
-from pytket_qir.qir import write_qir_file
+from pytket_qir.qir import write_qir_file  # type: ignore
```
cqc-alec: Is this (type ignore) necessary?

cqc-alec (changes requested):
Please also remove the wheels from the repo.

RolandMacDoland:
> Please also remove the wheels from the repo.

Sure. Done.

pytket_qir/gatesets/base.py
RolandMacDoland: Thanks.

tests/qir_test_files/WASM.ll
RolandMacDoland: I have been thinking about this a little. So far, `pyqir` requires passing a name to `SimpleModule` which will then be used to generate this line. I have no real idea for generic use case now (other than maybe using the circuit name if it exists) but in tests I can pass on the name of the test file to make it more explicit.

RolandMacDoland: Done.

pytket_qir/qir.py
RolandMacDoland: It is just a check for emptiness. But OK, I can be more explicit.

RolandMacDoland: Done.

pytket_qir/qir.py
RolandMacDoland: Done.

pytket_qir/qir.py
RolandMacDoland: I had the same reflection when I hardcoded this: it is not ideal. It'd probably be better if it was user defined.

pytket_qir/qir.py
RolandMacDoland: Thanks.

pytket_qir/qir.py
RolandMacDoland: Thanks.

pytket_qir/qir.py
RolandMacDoland: Thanks.

pytket_qir/qir.py
RolandMacDoland: True enough.
Done.

pytket_qir/qir.py
RolandMacDoland: I think you can have `in_width == 0` since you can define a `WASMOp` for which input and/or output registers are empty. However, I am still a bit confused about that assertion. Do I understand correctly that this assertion must say: " If there are empty registers (which is externally correct in the sense of the caller passing correct arguments), I (in the sense of that current piece of code) should be dealing with it but I am currently not."

RolandMacDoland: Done.

pytket_qir/gatesets/base.py
RolandMacDoland: Yes, this is one bit that I'd better find another solution. In brief, initially, `opnat`, `opname` and `opspec` were enums hence immutable. But it turned out while working on including WASM boxes that I needed mutability to define a `QirGate` object for it at runtime and especially the function name. This is why I re-define an enum to hold that name [here](https://github.com/CQCL/pytket-qir/blob/079202fe029c989171a622483421b73cd1e9319b/pytket_qir/qir.py#L488). Therefore, in `QirGate`, `opname` is defined as a generic `Enum` type. I'd be happy to hear another solution as I am not convinced this is the best one.

tests/conftest.py
RolandMacDoland: Hm, at some point, `mypy` on CI complained about it. Will double check. Apparently, it doesn't anymore.
Done.

pytket_qir/qir.py
RolandMacDoland: There is also a thread of discussion about it on Slack where Pete specifically mentioned WASM input and output ints being 64-bits. However, I recall previous discussions where 32-bits ints were put on the table. Probably best to clarify it.

RolandMacDoland: Done.

pytket_qir/qir.py
```diff
@@ -228,6 +242,29 @@ def block_to_circuit(self, block: QirBlock, circuit: Circuit) -> Circuit:
                 circuit.add_gate(op, unitids)
             elif instr.instr.is_qir_call:  # Setting the conditional bit for branching.
                 c_reg_index = instr.instr.call_func_params[0].constant.result_static_id
+            elif instr.instr.is_call:  # WASM external call.
+                if not self.wasm_handler:
```
cqc-alec:
```suggestion
                if self.wasm_handler is not None:
```
It's usually better to be explicit for this kind of thing: `if not x` may evaluate to false for all sorts of cases that you may or may not want to include: empty strings, empty sets, etc, as well as being user-definable. This avoids any ambiguity.

pytket_qir/utils/utils.py
```diff
@@ -0,0 +1,6 @@
+from enum import Enum
+
+
+class WASMType(Enum):
+    INT32 = 32
+    INT64 = 64
```
cqc-alec: This feels like a hack, attaching semantic significance to the enum values. (E.g. what if we want to add other 32-bit WASM types in the future?) Better not to rely on this in the code..

pytket_qir/qir.py
cqc-alec: Did you clarify it? 64 is still hard-coded here.

pytket_qir/qir.py
```diff
@@ -400,11 +471,76 @@ def condition_zero_block():
...
+            outputs: List[str] = []
+            for reglist, sizes in [
+                (inputs, op.input_widths),
+                (outputs, op.output_widths),
+            ]:
+                for in_width in sizes:
+                    assert in_width >= 0
+                    wasm_bits = args[:in_width]
+                    args = args[in_width:]
+                    regname = wasm_bits[0].reg_name
```
cqc-alec, line 499: This won't work if `in_width == 0`.

pytket_qir/gatesets/base.py
cqc-alec: I don't quite understand the problem is here. What goes wrong if you make `opname` a string?

cqc-alec (changes requested):

pytket_qir/qir.py
RolandMacDoland: Makes a lot of sense. Thanks.

pytket_qir/qir.py
Roland-djee: Yes, but isn't it exactly what we want ? Or am I getting confused ? We don't want to proceed any further if the WASMOp carries empty registers ?

pytket_qir/qir.py
RolandMacDoland: Oh yes sorry I missed those ones my bad. I will fix hat together with another solution for inputting wasm types sizes.

pytket_qir/qir.py
RolandMacDoland: True enough. I am a bit confused about it tbh. What was wrong with `in_width > 0` ? I guess it caught the right case ? Ie. we don't want to process further `WASMOp`s that carry over empty registers ?

pytket_qir/utils/utils.py
RolandMacDoland: OK, I can try to think of another solution here.

pytket_qir/qir.py
cqc-alec: As far as I know, a `WASMOp` with 0 input registers (but >0 output registers) is valid: it still performs some computation and outputs a (fixed) answer. If this is the case, we should handle it. But I haven't actually checked. If in fact it is invalid, we should throw a descriptive exception.

pytket_qir/gatesets/base.py
RolandMacDoland: Well, the issue is that WASM function calls are defined as external function calls in `pyqir-generator` (specifically [here](https://github.com/CQCL/pytket-qir/blob/9ed1c28417f44c8b741d55df23502fd1f8a5af60/pytket_qir/qir.py#L525)). To define them, I have to add ir to the `pyqir` module from an input gateset (teh machinery [here](https://github.com/CQCL/pytket-qir/blob/9ed1c28417f44c8b741d55df23502fd1f8a5af60/pytket_qir/qir.py#L376)). Previously, since the gateset was user defined and inputted alongside the circuit, it didn't need to be mutated at runtime. Now, when adding WASM calls, I need to update the gateset when processing the WASMOp, which means creating a new `QirGate` ([here](https://github.com/CQCL/pytket-qir/blob/9ed1c28417f44c8b741d55df23502fd1f8a5af60/pytket_qir/qir.py#L503)), updating the gateset ([here](https://github.com/CQCL/pytket-qir/blob/9ed1c28417f44c8b741d55df23502fd1f8a5af60/pytket_qir/qir.py#L509)) and finally updating the module ([here](https://github.com/CQCL/pytket-qir/blob/9ed1c28417f44c8b741d55df23502fd1f8a5af60/pytket_qir/qir.py#L518)). This was impossible with the previously defined datatypes (`OpName`, `OpType` etc) that were all enums. But in truth, the only one that needs to be updated is the `opname` field in `QirGate`. This is why I defined it as a generic `Enum` type here.

pytket_qir/qir.py
RolandMacDoland: Ah OK, thanks I see. Will do.

RolandMacDoland: OK after a quick check, it turns out that the only invalid WASM case is when there are no inputs not outputs, this means that `in_width == 0` should be handled.

RolandMacDoland: Done.

pytket_qir/qir.py
RolandMacDoland: Done.

pytket_qir/utils/utils.py
RolandMacDoland: Done.

pytket_qir/qir.py
```diff
@@ -400,11 +480,79 @@ def condition_zero_block():
...
+                    if wasm_bits != list(cregs[regname]):
+                        WASMError("WASM ops must act on entire registers.")
+                    reglist.append(regname)
+
+            try:
+                bit_reg = circ.get_c_register(inputs[0])
+            except RuntimeError:
+                bit_reg = BitRegister("wasm_input", 0)
```
cqc-alec: 1. Are you sure you would get `RuntimeError` if the index is out of range, not `IndexError`?
2. It would be good to add a test covering this edge case.
3. Will this still work if the circuit contains two zero-input WASM boxes? Won't there then be a clash of register names?
4. If the answer to (3) is yes, or if no and we have to have some unique register name, can we just use that solution in all cases instead of special-casing the 0 case?

pytket_qir/qir.py
```diff
@@ -87,14 +94,25 @@
...
+        wasm_handler: Optional[WasmFileHandler],
+        wasm_int_type: Optional[WasmInt] = WASMI32,
+    ) -> None:
         self.module = QirModule(file_path)
-        self.gateset: Union[GateSet, CustomGateSet] = (
-            gateset if gateset else PYQIR_GATES
-        )
+        self.gateset: CustomGateSet = gateset if gateset else PYQIR_GATES
+        self.wasm_handler = wasm_handler
+        assert wasm_int_type
```
cqc-alec:
```suggestion
        assert wasm_int_type is not None
```

pytket_qir/gatesets/base.py
cqc-alec: Can we have a specific enum of op types, which includes the WASM op? It would be more satisfying if this were captured as a single enum rather than a generic enum type, which could be anything at all.

pytket_qir/qir.py
```diff
@@ -228,6 +245,33 @@ def block_to_circuit(self, block: QirBlock, circuit: Circuit) -> Circuit:
...
+                if matched_str is None:
+                    raise ValueError(
+                        "The WASM function call name is not propely defined."
+                    )
+                # WASM function call parameters.
+                param_regs = []
+                for c_reg_index in range(len(instr.func_args)):
+                    c_reg_name = "c_reg_wasm" + str(c_reg_index)
+                    param_regs.append(
+                        circuit.add_c_register(c_reg_name, self.wasm_int_type.size)
```
cqc-alec: There are a couple of other places in this file where 64 seems to be hard-coded: lines 226, 227 and 295.

cqc-alec (changes requested):

pytket_qir/qir.py
RolandMacDoland: 1. Is not an out-of-range error but a non-existent name. At the very least, `inputs` will be empty.
2. Certainly. I can do that.
3. This register is only created for the purposes of initialising the ssa variable to 0 when there is no input to the WASMOp. I don't think it really matters if there are multiple zero-input WASM boxes. And it is not reused anywhere else.
4. I don't think we need a unique register here.

pytket_qir/qir.py
RolandMacDoland: Yes but those are different. There are to hold constants as llvm int64. Maybe worth using another data type to make it more explicit.

RolandMacDoland: Done.

pytket_qir/gatesets/base.py
RolandMacDoland: OK let me experiment with that idea.

RolandMacDoland: I came up with a slightly better solution but I couldn't get away with creating a singleton enum on the fly for the purposes of holding the WASM Function name. This is because all the machinery for defining external function calls rely on enums syntax and it is really tedious to adapt that for a specialised case where the gate definition will be of another (more flexible?) type.

RolandMacDoland: Well, no more luck. To comply with mypy, I have to define a union of types with the generic `Enum`. I don't really see any way around this, sorry.

pytket_qir/qir.py
cqc-alec: Please add a test case including two WASM ops with 0 inputs.

cqc-alec (changes requested):

pytket_qir/qir.py
RolandMacDoland: You're right as always. Done.

tests/qir_test_files/WASM_noinputs.ll
```diff
@@ -0,0 +1,16 @@
+source_filename = "Generated from input pytket circuit and wasm_adder.wasm file."
+
+declare i32 @__quantum__hybrid__add_one__body(i32)
```
cqc-alec, line 3: If this is a no-inputs function shouldn't it be `()` rather than `(i32)` in the QIR?

cqc-alec (changes requested):

tests/qir_test_files/WASM_noinputs.ll
RolandMacDoland: Not sure. We explicitly made it such that if there isn't any input in the WASMOp, then the SSA variable input  in QIR Is initialised to 0.

cqc-alec (approved):

PR #7, opened 2022-09-15
Arithmetic bitwise in generator

RolandMacDoland:
This PR addresses the addition of classical arithmetic and bitwise operators in the new QIR generator class.

PR #8, opened 2022-09-21
Reg2const arithmetic and tracking previously set ssa vars

RolandMacDoland:
This PR addresses the addition of two functionalities:

- `reg2const` arithmetic: this requires to keep track of registers set with a constant value through the `SetBitsOp` for potential later reuse.
- Keep track of ssa variables: in the same line as above, there is now functionality to keep track of previously created ssa vars for later reuse.

This work also encompasses an architectural re-design of the package in three main parts:

- A QIR parser
- A QIR generator
- A module that extends the `pyqir` `SimpleModule` to handle custom define quantum functions.


RolandMacDoland:
Hi @cqc-alec , I recall Pete was also requesting for reviewing this PR. Since he is busy with IEEE this week, I will request his review next week.

cqc-alec (commented):
Thanks. As with the other PR, I'm not sure there's much point my reviewing this until it can be tested cross-platform with public pyqir wheels.

As a general point I would like to see more tests demonstrating correctness of the conversions on (small) whole programs.

bamarsha:
PyQIR 0.6.0 is released on PyPI now.

RolandMacDoland:
> PyQIR 0.6.0 is released on PyPI now.

Thanks @samarsha this is going to be helpful.

cqc-alec:
Since this includes changes from #6 I will review once that is merged.

cqc-alec (changes requested):
Needs merge from `develop` please.

RolandMacDoland:
> Needs merge from `develop` please.

Yes, sure. But since this PR contains changes from #7, I'd rather merge the other first if that's possible.

tests/qir_test.py
```diff
@@ -596,6 +608,7 @@ def test_generate_wasmop(self) -> None:
         ir_bytes = cast(bytes, circuit_to_qir(circuit, wasm_path=wasm_file_path))
 
         ll = bitcode_to_ir(ir_bytes)
+        # print(ll)
```
cqc-alec: Delete this line.

pytket_qir/generator.py
```diff
@@ -0,0 +1,407 @@
...
+            if op.get_name() == "AND":
+                optype = BitWiseOp.AND
+            elif op.get_name() == "OR":
+                optype = BitWiseOp.OR
+            elif op.get_name() == "XOR":
+                optype = BitWiseOp.XOR
+        else:
+            params = op.params
+            if optype == OpType.TK1:
+                params = [op.params[1], op.params[0] - 0.5, op.params[2] + 0.5]
```
cqc-alec: Why this change of params?

pytket_qir/generator.py
```diff
@@ -0,0 +1,407 @@
...
+
+    def _to_qis_bits(self, args: List[Bit]) -> Sequence[Result]:
+        if args:
+            return [self.module.module.results[bit.index[0]] for bit in args[:-1]]
+        return []
+
+    def _reg2ssa_var(self, bit_reg: BitRegister, int_size: int) -> Value:
+        # A utility function to convert from a pytket
+        #  BitRegister to an SSA variable via pyqir types.
+        reg_name = bit_reg[0].reg_name
```
cqc-alec, line 157: Should we handle empty registers?

pytket_qir/generator.py
```diff
@@ -0,0 +1,407 @@
...
+        elif isinstance(op, SetBitsOp):
+            for reglist, sizes in [
+                (inputs, [op.n_inputs]),
+                (outputs, [op.n_outputs]),
+            ]:
+                for in_width in sizes:
+                    if in_width > 0:
+                        com_bits = args[:in_width]
+                        args = args[in_width:]
+                        regname = com_bits[0].reg_name
```
cqc-alec: What if `in_width` is zero? Should handle edge case.

pytket_qir/module.py
```diff
@@ -0,0 +1,70 @@
...
+        name: str,
+        num_qubits: int,
+        num_results: int,
+        gateset: Optional[CustomGateSet] = None,
+        wasm_handler: Optional[WasmFileHandler] = None,
+    ) -> None:
+        self.module = SimpleModule(name, num_qubits, num_results)
+        self.builder = self.module.builder
+        self.qis = BasicQisBuilder(self.builder)
+        self.gateset = gateset if gateset else PYQIR_GATES
```
cqc-alec, line 47:
```suggestion
        self.gateset = gateset if gateset is not None else PYQIR_GATES
```

pytket_qir/module.py
```diff
@@ -0,0 +1,70 @@
...
+        self.wasm_handler = wasm_handler
+
+    @property
+    def gateset(self):
+        """A getter for the gateset."""
+        assert self._gateset
```
cqc-alec:
```suggestion
        assert self._gateset is nor None
```

pytket_qir/parser.py
```diff
@@ -0,0 +1,370 @@
...
+                    # and add it to the register map.
+                    output_name = cast(str, instr.output_name)
+                    c_reg_name3 = "%" + output_name
+                    c_reg3 = circuit.add_c_register(
+                        c_reg_name3, self.qir_int_type.width
+                    )  # Int64 supported in LLVM/QIR and L3.
+                    c_reg_map[3] = c_reg3
+                    # import pdb; pdb.set_trace()
+                    # instr = cast(instr, PyQirInstruction)
+                    # assert isinstance(instr, PyQirInstruction)
```
cqc-alec: Delete.

cqc-alec (changes requested):
- Please remove the 4 wheels.
- Please remove `utils.py~`.

RolandMacDoland:
> * Please remove the 4 wheels.
> * Please remove `utils.py~`.

Thanks, I don't know why the references to these files keep propagating through various PRs although I've removed them in several other places.

tests/qir_test.py
RolandMacDoland: Thanks. Done.

pytket_qir/generator.py
RolandMacDoland: Not sure. The only calls to that method are [here](https://github.com/CQCL/pytket-qir/blob/21acfb115093a45f6dc3dcd732d070f021e9167a/pytket_qir/generator.py#L297) and [here](https://github.com/CQCL/pytket-qir/blob/21acfb115093a45f6dc3dcd732d070f021e9167a/pytket_qir/generator.py#L308) which are guaranteed to pass non-empty register.

pytket_qir/module.py
RolandMacDoland: Thanks.

pytket_qir/module.py
RolandMacDoland: Thanks

pytket_qir/parser.py
RolandMacDoland: Sorry, left over from debugging. Done.

pytket_qir/generator.py
RolandMacDoland: When I create a `SetBitsOp` with an empty register, I get a runtime error:

```python
>>> c = Circuit()
>>> b = c.add_c_register("b", 0)
>>> c.add_c_setreg(2, b)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: An operation must act on at least one wire
```

This is possible though when properly creating a `SetBitsOp` 

```python
>>> c = Circuit()
>>> b = c.add_c_register("b", 3)
>>> c.add_c_setreg(2, b)
[SetBits(010) b[0], b[1], b[2]; ]
>>> com  = c.get_commands()[0]
>>> com.op
SetBits(010)
>>> com.op.n_inputs
0
>>> com.op.n_outputs
3
```

But in all cases, `n_inputs` is always 0. In this case, we are only interested in the output register which is always set otherwise we get the error above.

pytket_qir/gatesets/pyqir/pyqir.py
```diff
@@ -42,6 +42,7 @@
 PYQIR_GATES = CustomGateSet(
     name="PyQir",
     template=Template("__quantum__${opnat}__${opname}__${opspec}"),
+    base_gateset=_TK_TO_PYQIR.keys() | set(),
```
cqc-alec:
```suggestion
    base_gateset=set(_TK_TO_PYQIR.keys()),
```
Simpler, and makes it always a `set`.

pytket_qir/generator.py
```diff
@@ -0,0 +1,428 @@
...
+        wasm_int_type: types.Int = types.Int(32),
+        qir_int_type: types.Int = types.Int(64),
+    ) -> None:
+        self.circuit = circuit
+        self.module = module
+        self.wasm_int_type = wasm_int_type
+        self.qir_int_type = qir_int_type
+        self.cregs = _retrieve_registers(self.circuit.bits, BitRegister)
+        self.target_gateset = self.module.gateset.base_gateset
+        self.rebase_to_gateset = auto_rebase_pass(self.target_gateset)
```
cqc-alec, line 114: This may throw an exception if `auto_rebase_pass` doesn't handle the base gateset. I guess this is the best we can do, but we should probably document it somewhere.

tests/conftest.py
```diff
@@ -260,6 +286,16 @@ def file_name() -> str:
     return "SimpleCircuit.ll"
 
 
+@fixture
+def rebased_circuit() -> Generator:
```
cqc-alec: Does this need to be a fixture. as it's only used in one test?

pytket_qir/generator.py
cqc-alec: Agree about the first; not quite sure about the second: can we assume a `ClassicalExpBox` has no classical wires?

pytket_qir/generator.py
cqc-alec: Have you tried with latest pytket? This error was removed in https://github.com/CQCL/tket/pull/589 .

cqc-alec (changes requested):

pytket_qir/gatesets/pyqir/pyqir.py
RolandMacDoland: Thanks, but why always ? The pipe operator won't work if the `keys()` array is empty ?

cqc-alec: > Thanks, but why always ? The pipe operator won't work if the `keys()` array is empty ?

I don't think we need the pipe operator? `keys()` does not return a `set` so with your version we get different types depending if it is empty or not.

pytket_qir/generator.py
RolandMacDoland: I am aware of the fragility of that approach. Previously, it raised an error if the gate wasn't in the translation dictionary. However, there could be scenarii (I think) where a user defines a `pytket` circuit and applies optimisations that introduce `tket` specific gates (TK1). Correct me if I am utterly wrong in that assumption though. However, one thing that you can bear in mind is that the default `pyqir` gateset is probably "standard enough" to accommodate rebases from `pytket`.  But OK I can comment on that of course, aside to letting the `NoAutoRebase` going through.

pytket_qir/gatesets/pyqir/pyqir.py
RolandMacDoland: Oh really ? Hm OK thanks for the clarification then.

pytket_qir/generator.py
RolandMacDoland: Done.

tests/conftest.py
RolandMacDoland: Yes, that is probably a bad habit. But I try to put as much as I can in fixtures as not to clutter tests too much. Esspecially the `arrange` part.

RolandMacDoland: Done.

pytket_qir/generator.py
RolandMacDoland: It looks impossible to add a `ClassicalExpBox` with empty registers:

```python
>>> c = Circuit(0)
>>> a = c.add_c_register("a", 0)
>>> b = c.add_c_register("b", 0)
>>> c.add_classicalexpbox_register(a ^ b, b)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: An operation must act on at least one wire
```

And also: the call is within a loop that gets exectuted iff there are inputs ?

cqc-alec: > It looks impossible to add a `ClassicalExpBox` with empty registers:

With pytket==1.8.0rc0 it is not impossible (this was recently changed):
```
In [2]: from pytket import Circuit

In [3]: c = Circuit(0)

In [4]: a = c.add_c_register("a", 0)

In [5]: b = c.add_c_register("b", 0)

In [6]: c.add_classicalexpbox_register(a ^ b, b)
Out[6]: [ClassicalExpBox; ]
```

RolandMacDoland: Yes, I tried that out just before upgrading to the latest version (related to the next comment).

pytket_qir/generator.py
RolandMacDoland: Ok so now, it makes sense to throw an exception ? If `in_width  = 0` is valid `SetBits` that means the user is trying to assign a value to an empty register which is very likely not what it intends to do.

cqc-alec: > Ok so now, it makes sense to throw an exception ? If `in_width = 0` is valid `SetBits` that means the user is trying to assign a value to an empty register which is very likely not what it intends to do.

I don't believe we should throw an exception just because it seems like an odd thing to do. By doing so, we may force users to insert special-case handling in their own code, which will be an annoyance for them. Edge cases should be handled painlessly, if possible.

RolandMacDoland: Yes but that can be the sign of a genuine bug ? I.e. an assignment to a wrong register ?

cqc-alec: (Remember that "the user" here could be some generated code.)
Basically, if it isn't easy to handle, fine, but I'd like to understand why -- since that would probably indicate some poor design choices upstream...

RolandMacDoland: Upstream where ? In the user process ? At the `SetBits` op level ?
Furthermore, I don't see a clear way to handle that case painlessly without further understanding of the user's intentions which looks like a complicated thing to do. For instance, setting the register size to an arbitrary value if empty ? I am not too sure it makes sense. But maybe I am wrong.

cqc-alec: > Upstream where ? In the user process ? At the `SetBits` op level ? Furthermore, I don't see a clear way to handle that case painlessly without further understanding of the user's intentions which looks like a complicated thing to do. For instance, setting the register size to an arbitrary value if empty ? I am not too sure it makes sense. But maybe I am wrong.

By "upstream" I mean in some code that this depends on: that may be elsewhere in pytket-qir or it may be in pytket or it may be in pyqir. But the user can reasonably expect to be able to do this, whether or not it seems sensible. The semantics of `SetBits` with no wires are unambiguous: it is a no-op. But sometimes code generation can emit no-ops and they shouldn't cause exceptions. I'm not sure why we need to understand the user's intentions beyond this? Perhaps I am missing something.

RolandMacDoland: Thanks for this. It is clear. What I meant is how do you disambiguate between a code generation that emits a no-op as part of a process and a genuine bug ? If you could, it would justify to handle the cases appropriately: ie raising an exception or painlessly going through. If not, how do you handle the case other than a generic behaviour like raising an exception ?

cqc-alec: > Thanks for this. It is clear. What I meant is how do you disambiguate between a code generation that emits a no-op as part of a process and a genuine bug ? If you could, it would justify to handle the cases appropriately: ie raising an exception or painlessly going through. If not, how do you handle the case other than a generic behaviour like raising an exception ?

We can't, and it's not our jpb to distinguish it -- any more than it is the job of gcc to error if a line of C code has no effect. We can optimize it away if we want -- but should allow it. If it's a bug, that's someone else's problem. Am I missing the point?

RolandMacDoland: No, I think we're on the same page. But in this context, since setting a `SetBits` with empty registers is now valid, it could still be a potential bug and I don't see any other way to handle the case without raising an exception if I am to strive for correctness ?

pytket_qir/generator.py
RolandMacDoland: Done.

pytket_qir/generator.py
RolandMacDoland: Done.

cqc-alec (approved):
Could you please create a (low-priority) task to make `SetBits` and `ClassicalExpBox` support the trivial no-wires case? Ideally we should support this but maybe there are issues with pytket that prevent it.

RolandMacDoland:
> Could you please create a (low-priority) task to make `SetBits` and `ClassicalExpBox` support the trivial no-wires case? Ideally we should support this but maybe there are issues with pytket that prevent it.

Sure. But don't they already support it ? I thought from the examples above that setting them with empty registers was valid. Or, am I getting confused ?

Thanks for the review.

cqc-alec:
> > Could you please create a (low-priority) task to make `SetBits` and `ClassicalExpBox` support the trivial no-wires case? Ideally we should support this but maybe there are issues with pytket that prevent it.
> 
> Sure. But don't they already support it ? I thought from the examples above that setting them with empty registers was valid. Or, am I getting confused ?

E.g. https://github.com/CQCL/pytket-qir/pull/8/commits/e0c46daf5c66f8cee80f27177d1663c3bc98e8d5

I still don't get why we have to raise an exception here (I mean, I get it in terms of how the code works currently, but I don't see why it has to be that way).

Anyway this is low-priority and shouldn't block merging this!

RolandMacDoland:
> > > Could you please create a (low-priority) task to make `SetBits` and `ClassicalExpBox` support the trivial no-wires case? Ideally we should support this but maybe there are issues with pytket that prevent it.
> > 
> > 
> > Sure. But don't they already support it ? I thought from the examples above that setting them with empty registers was valid. Or, am I getting confused ?
> 
> E.g. [e0c46da](https://github.com/CQCL/pytket-qir/commit/e0c46daf5c66f8cee80f27177d1663c3bc98e8d5)
> 
> I still don't get why we have to raise an exception here (I mean, I get it in terms of how the code works currently, but I don't see why it has to be that way).
> 
> Anyway this is low-priority and shouldn't block merging this!

Tbf, I don't understand very clearly either what you are expecting. I am a little confused between the new behaviour of the ops and handling it in the context of correctness here. Maybe best to discuss that in Slack ?

PR #9, opened 2022-10-12
Wasm calls

RolandMacDoland:
This PR addresses the update to new `pyqir` and `pytket` releases. Especially around the generation of function attributes in `pyqir` and WASM function checks in `pytket`. This last point addresses the previous mismatch when a `WASMOp` had empty register inputs, the SSA variable was initialised to 0. Now, the `WASM` function isn't passed any argument.


tests/qir_test.py
```diff
@@ -557,6 +566,41 @@ def test_bitwise_ops(self, circuit_bitwise_ops: Circuit) -> None:
...
+        c1 = circuit.add_c_register("c1", 64)
+
+        circuit.add_wasm_to_reg("add_one", wasm_handler, [c0], [c1])
+        ir_bytes = circuit_to_qir_bytes(circuit, wasm_path=wasm_file_path)
+
+        ll = bitcode_to_ir(ir_bytes)
+        assert ll in exp_data
+
+    def test_generate_wasmop_with_empty_inputs(self) -> None:
+        wasm_file_path = qir_files_dir / "wasm_adder.wasm"
```
cqc-alec: Should this be `wasm_empty_adder.wasm`? I see that file has been added but it is not referenced anywhere.

tests/qir_test_files/WASM_noinputs.ll
```diff
@@ -0,0 +1,16 @@
+source_filename = "Generated from input pytket circuit and wasm_adder.wasm file."
```
cqc-alec: According to this line it was generated from the same file as `WASM.ll`. Is that right?

tests/qir_test.py
```diff
@@ -104,6 +106,13 @@ def test_grover_circuit(self, grover_circuit: Circuit) -> None:
         circuit = circuit_from_qir(grover_bc_file_path)
         assert circuit == grover_circuit
 
+    def test_wasm_only(self) -> None:
+        wasm_only_bc_file_path = qir_files_dir / "wasm_only_test.bc"
+        wasm_file_path = qir_files_dir / "wasm_file.wasm"
+        wasm_handler = WasmFileHandler(str(wasm_file_path))
+        circuit = circuit_from_qir(wasm_only_bc_file_path, wasm_handler=wasm_handler)
+        assert circuit.depth() == 1
```
cqc-alec, line 117: Can we assert more than this, i.e. what the commands are?

pytket_qir/qir.py
```diff
@@ -87,14 +93,28 @@
...
+    ) -> None:
         self.module = QirModule(file_path)
-        self.gateset: Union[GateSet, CustomGateSet] = (
-            gateset if gateset else PYQIR_GATES
-        )
+        self.gateset: CustomGateSet = gateset if gateset else PYQIR_GATES
+        self.wasm_handler = wasm_handler
+        assert wasm_int_type is not None
+        self.wasm_int_type = wasm_int_type
+        assert qir_int_type
```
cqc-alec:
```suggestion
        assert qir_int_type is not None
```

pytket_qir/qir.py
```diff
@@ -87,14 +93,28 @@
...
+        wasm_int_type: Optional[types.Int] = types.Int(32),
+        qir_int_type: Optional[types.Int] = types.Int(64),
+    ) -> None:
         self.module = QirModule(file_path)
-        self.gateset: Union[GateSet, CustomGateSet] = (
-            gateset if gateset else PYQIR_GATES
-        )
+        self.gateset: CustomGateSet = gateset if gateset else PYQIR_GATES
+        self.wasm_handler = wasm_handler
+        assert wasm_int_type is not None
```
cqc-alec: For both `wasm_int_type` and `qir_int_type`, we cannot assert that they are not `None` if they are declared as `Optional`. If we don't want to allow `None` they should not be `Optional`.

pytket_qir/qir.py
```diff
@@ -400,11 +482,78 @@ def condition_zero_block():
...
+            outputs: List[str] = []
+            for reglist, sizes in [
+                (inputs, op.input_widths),
+                (outputs, op.output_widths),
+            ]:
+                for in_width in sizes:
+                    assert in_width >= 0
+                    wasm_bits = args[:in_width]
+                    args = args[in_width:]
+                    regname = wasm_bits[0].reg_name
```
cqc-alec: This will fail if `in_width` is zero, how do we handle that?

cqc-alec (changes requested):

tests/qir_test.py
RolandMacDoland: Sorry, I forgot to commit these changes. Done.

tests/qir_test_files/WASM_noinputs.ll
RolandMacDoland: Also forgot to commit. Done.

tests/qir_test.py
RolandMacDoland: Of course. Done.

pytket_qir/qir.py
RolandMacDoland: Sure but would it make sense to have `wasm_int_type` as non-optional but defaulted parameter and `wasm_handler` still being optional ? But then that would mean rearranging the parameter list and potential API breaks. Let me see if this could be made with minimal disruptive changes.

RolandMacDoland: Done.

pytket_qir/qir.py
RolandMacDoland: Done.

pytket_qir/qir.py
RolandMacDoland: Actually there is no need for this assertion because the corner case is that if `size` is empty. In that case, this loop is not executed and `reglist` (therefore `inputs`) is empty. This is handled later in the `try/except` block here:

```python
try:
                bit_reg = circ.get_c_register(inputs[0])
                input_type_list = [types.Int(wasm_int_size)]
 except IndexError:
                input_type_list = []

```

RolandMacDoland: And this assertion is removed in a later PR ;).

pytket_qir/qir.py
```diff
@@ -103,17 +103,15 @@ class QirParser:
     def __init__(
         self,
         file_path: str,
-        gateset: Optional[CustomGateSet],
-        wasm_handler: Optional[WasmFileHandler],
-        wasm_int_type: Optional[types.Int] = types.Int(32),
-        qir_int_type: Optional[types.Int] = types.Int(64),
+        gateset: CustomGateSet = None,
+        wasm_handler: WasmFileHandler = None,
```
cqc-alec: Since `None` isn't a valid value for `CustomGateSet` or `WasmFileHandler`, these type annotations don't make sense. If we want `None` to be a default then they should be `Optional[...]`.

cqc-alec (changes requested):

pytket_qir/qir.py
RolandMacDoland: There is a thing here, for instance, I tried to default `gateset` to `PYQIR_GATES` but the default value didn't seem to be propagated further the initialiser. I.e. `gateset` was `NoneType` even if defaulted to `PYQIR_GATES`. Unless I did something egregiously wrong, my understanding is that (at least in that case) handling a default value works if it is set to `None` ? 

Ah, OK I see what you mean: `None` isn't a valid `CustomGateSet` type. Hm, I get it better but nonetheless, I couldn't default it to `PYQIR_GATES` as I attempted above. Other point, why `mypy` doesn't pick up these non-sensical type situations ? 

Let me dig in that a little more.

RolandMacDoland: Done. This is still weird: I can't default `gateset` tp `PYQIR_GATE` as I would like to, so making it `Optional`.

cqc-alec (approved):

PR #10, opened 2022-10-14
Support rt record functions, select and zext in parser

RolandMacDoland:
This PR addresses the support for tagged or untagged runtime functions to record results from the device. Specifically:

- `int_record`
- `bool_record`
- `result_record`

Besides, it also allows for the support of `select` and `zext` gates.


tests/qir_test.py
```diff
@@ -118,6 +117,126 @@ def test_wasm_only(self) -> None:
...
+                com.op.type == OpType.Conditional
+            ):  # Raising an error if I use .commands_of_type(OpType.Conditional)
+                conditionals.append(com)
+
+        conditional0 = conditionals[0]
+        assert conditional0.op.type == OpType.Conditional
+        assert conditional0.bits == list(output_register)
+        assert conditional0.op.value == 1
+        assert (
+            sum([int(n) * 2**k for k, n in enumerate(conditional0.op.op.values)])
```
cqc-alec: Why is the cast to `int` necessary?

tests/qir_test.py
```diff
@@ -118,6 +117,126 @@ def test_wasm_only(self) -> None:
...
+            == 99
+        )
+
+        conditional1 = conditionals[1]
+        assert conditional1.op.type == OpType.Conditional
+        assert conditional1.bits == list(output_register)
+        assert conditional1.op.value == 0
+        assert (
+            sum([int(n) * 2**k for k, n in enumerate(conditional1.op.op.values)])
```
cqc-alec: ... and here?

pytket_qir/gatesets/base.py
```diff
@@ -38,7 +38,6 @@ class OpName(Enum):
     AND = "and"
     OR = "or"
     XOR = "xor"
-    WASM = "wasm"
```
cqc-alec, line 41: Is this not needed anymore?

tests/qir_test.py
RolandMacDoland: Reminiscent of the old debugging times. Removed.

tests/qir_test.py
RolandMacDoland: Removed.

pytket_qir/gatesets/base.py
RolandMacDoland: No. This name is now dynamically created [here](https://github.com/CQCL/pytket-qir/blob/90b44b57a05a0b3f585b901148f2b9fed13c80e0/pytket_qir/generator.py#L286).

cqc-alec (approved):

PR #15, opened 2022-11-07
Fix ssa tracking in parser

RolandMacDoland:
This PR addresses the addition of logging functionality for ssa variables (registers) created and used in branches operations.  It also checks for cases where 

- classical arithmetic
- wasm calls 
- `select`

are used in branches where inputs are defined outside the scope of the branch. 

Lastly, it provides support for the `__quantum__qir__read_result` function.


pytket_qir/parser.py
```diff
@@ -326,35 +337,64 @@ def block_to_circuit(self, block: QirBlock, circuit: Circuit) -> Circuit:
...
+                    circuit.add_c_setreg(value, c_reg)
+                elif func_name.startswith("__quantum"):
+                    matched_str = re.search("__quantum__(.+?)__(.+?)__(.+)", func_name)
+                    if matched_str is None:
+                        raise WASMError(
+                            "The WASM function call name is not properly defined."
+                        )
+                    # WASM function call parameters.
+                    param_regs = []
+                    instr_arg = instr.func_args[0]
```
cqc-alec: Can `instr.func_args` be empty?

tests/qir_test.py
```diff
@@ -240,23 +240,32 @@ def test_single_conditional(self, one_conditional_circuit: Circuit) -> None:
...
+        assert inverse_reg_map[condition_com.args[0]].reg_name == "%0"
+
         condition_circuit = condition_com.op.op.get_circuit()
 
         exp_condition_com = one_conditional_circuit.get_commands()[15]
         exp_condition_circuit = exp_condition_com.op.op.get_circuit()
 
-        assert condition_circuit == exp_condition_circuit
+        # Comparing commands as I get a type error while comparing circuits.
+        # TypeError: unhashable type: 'instancemethod'.
```
cqc-alec, line 263: Do you know why? Is this a bug in pytket?

tests/qir_test.py
```diff
@@ -278,26 +287,30 @@ def test_multiple_successive_conditionals(
...
         condition_circuit_1 = condition_com_1.op.op.get_circuit()
         condition_circuit_1_coms = condition_circuit_1.get_commands()
 
         exp_condition_com_1 = multiple_conditionals_circuit.get_commands()[10]
         exp_condition_circuit_1 = exp_condition_com_1.op.op.get_circuit()
         exp_condition_circuit_1_coms = exp_condition_circuit_1.get_commands()
 
-        non_condition_circuit_1 = Circuit(6, 6)
-        exp_non_condition_circuit_1 = Circuit(6, 6)
+        non_condition_circuit_1 = Circuit(6, 262)
```
cqc-alec, line 311: Is this blow-up in circuit size (from 6 to 262) necessary?

pytket_qir/parser.py
RolandMacDoland: OK might be worth raising an exception in case. Done.

tests/qir_test.py
RolandMacDoland: Not too sure. If I run the assertion on circuit equality, I have got the following error:

```sh
>       assert condition_circuit == exp_condition_circuit
E       AssertionError: assert [Reset q[0]; ... Reset q[0]; ] == [Reset q[0]; ... Reset q[0]; ]
E         (pytest_assertion plugin: representation of details failed: /Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/pprint.py:176: TypeError: unhashable type: 'instancemethod'.
E          Probably an object has a faulty __repr__.)

qir_test.py:269: AssertionError
```
However,  the assertion works in the debugger and returns `False` (although circuits are seemingly identical):

```sh
(Pdb) assert condition_circuit == exp_condition_circuit
*** AssertionError
(Pdb) condition_circuit == exp_condition_circuit
False
(Pdb)
```
But it succeeds if I compare commands explicitly.

tests/qir_test.py
RolandMacDoland: Yes, otherwise I get the error that the circuit doesn't hold enough unitids.

```sh
        for com in condition_circuit_1_coms:
            if not isinstance(com.op, Conditional):
>               non_condition_circuit_1.add_gate(com.op, com.args)
E               RuntimeError: Circuit does not contain unit with id: c[257]
```

pytket_qir/parser.py
```diff
@@ -326,35 +341,67 @@ def block_to_circuit(self, block: QirBlock, circuit: Circuit) -> Circuit:
...
+                    circuit.add_c_setreg(value, c_reg)
+                elif func_name.startswith("__quantum"):
+                    matched_str = re.search("__quantum__(.+?)__(.+?)__(.+)", func_name)
+                    if matched_str is None:
+                        raise WASMError(
+                            "The WASM function call name is not properly defined."
+                        )
+                    # WASM function call parameters.
+                    param_regs = []
+                    if instr.func_args is None:
```
cqc-alec: Can it be the empty list?

RolandMacDoland: Yep, that is silly. Thanks. Done.

cqc-alec (approved):

PR #16, opened 2022-11-14
Support rt record functions

RolandMacDoland:
This PR addresses the support for untagged runtime functions namely:

- `integer_record_output`
- `bool_record_output`
- `result_record_output`

and support for `read_result` instructions and few other bits like passing constant variables to functions.


cqc-alec:
Does this require the binary wheel you have added in this PR?

.github/workflows/test.yml
```diff
@@ -35,4 +35,4 @@ jobs:
       - name: Run tests
         working-directory: ./tests
         run: |
-          poetry run pytest qir_test.py
+          poetry run pytest parser_test.py generator_test.py
```
cqc-alec, line 38: Have all the tests for `qir_test.py` moved into `parser_test.py` and `generator_test.py`? If so, `qir_test.py` should be deleted; if not, it should be run!

pytket_qir/gatesets/base.py
```diff
@@ -38,13 +40,18 @@ class OpName(Enum):
     AND = "and"
     OR = "or"
     XOR = "xor"
+    INT = "integer"
+    BOOL = "bool"
+    RES = "result"
```
cqc-alec, line 45: These look like types rather than ops?

tests/conftest.py
```diff
@@ -19,6 +19,7 @@
 
 from pytest import fixture  # type: ignore
 
+# from pyqir import Builder, IntPredicate, Value  # type: ignore
```
cqc-alec: Remove commented line.

.github/workflows/test.yml
cqc-alec: Ah, I see it has been renamed, nm.

cqc-alec (changes requested):
Generally seems good, question about the wheel.

RolandMacDoland:
> Does this require the binary wheel you have added in this PR?

No, that's right. Removed. Thanks.

tests/conftest.py
RolandMacDoland: Thanks. Done.

pytket_qir/gatesets/base.py
RolandMacDoland: That's right. Maybe I should rename these datatypes something like:: `FuncName`, `FuncSpec` etc ?

RolandMacDoland: Done.

cqc-alec (approved):

Issue #18, opened 2022-11-24
Reintroduce flattening registers to account for non-simple circuits in CircBox


RolandMacDoland:
When translating from QIR nested conditions to pytket circuits, flattening registers must happen to avoid non-simplicity of circuits when boxed in condition `CircBox`.

This must happen and update the `self.set_creg` data member in `QirParser` that keeps log of locally created SSA variables for reuse.

PR #19, opened 2022-12-14
Handle conditionals

RolandMacDoland:
This PR addresses the correct handle of conditionals and goes beyond #18. 

After quite some meandering, it is a major refactor and implementation of a new class:  `QirConverter`. Its main responsibility is to create and transform the CFG from a QIR program to handle conditionals appropriately. In short, the `QirConverter` does:

- Create and topologically sort the CFG.
- Apply (or not) a simple optimisation that is to contract linear chains of blocks that are linked by jump instructions.
- Iterate through the CFG blocks and calling the `QirParser` to generate the corresponding circuit. Previous conditionality management in the `QirParser` now encumbers the `QirConverter`. The `QirParser` is solely responsible for returning a circuit from a given block.
- Guard the blocks appropriately based on Pete's (or John Gavin's ?) algorithm.
- Generating a final circuit where each QIR block is guarded.

Minor new helpful functionalities:

- Maintaining the CFG structure by considering `then` clauses as resuming the flow of the main circuit.
- Treating jumps as flow continuation.
- Extending the canonical `c` register for created ssa variables to avoid non-simplicity of circuits.
- Representing the `read_result` instruction as ` CopyBitOp`. 
- Keeping and returning a log of ssa variables as a circuit member `ssa_vars`.

It also adds another validation test from the RUS example file.


cqc-alec:
> considering else clauses as resuming the flow of the main circuit

I don't quite understand this. Wouldn't the "else" circuitry then be executed unconditionally?

RolandMacDoland:
> > considering else clauses as resuming the flow of the main circuit
> 
> I don't quite understand this. Wouldn't the "else" circuitry then be executed unconditionally?

Yes, exactly. If I condition the else circuit (which I did over the last week) I lose the domination tree structure for ssa instructions and I won't be able to generate any QIR. For all circuit examples I have seen so far, the output of true condition branches join the else ones. I know this is not an argument strong enough but if we stick to bits prepared in the 0 state initially, the else clause is the normal instruction flow.

cqc-alec:
So it seems that in all the QIR test cases here, the `else` clause actually does nothing. For example, in `SimpleConditionalCircuit.ll`, we have the following:

```
define void @main() #0 {
entry:
  call void @__quantum__qis__x__body(%Qubit* null)
  call void @__quantum__qis__x__body(%Qubit* inttoptr (i64 1 to %Qubit*))
  %0 = call i1 @__quantum__qis__read_result__body(%Result* null)
  br i1 %0, label %then, label %else

then:                                             ; preds = %entry
  call void @__quantum__qis__h__body(%Qubit* null)
  call void @__quantum__qis__h__body(%Qubit* inttoptr (i64 1 to %Qubit*))
  br label %continue

else:                                             ; preds = %entry
  br label %continue

continue:                                         ; preds = %else, %then
```
etc.

But what if instead of

```
else:                                             ; preds = %entry
  br label %continue
```

we had something like

```
else:                                             ; preds = %entry
  call void @__quantum__qis__x__body(%Qubit* null)
  br label %continue
```

In that case, the X gate should be applied _if and only if_ the condition in 

```
  br i1 %0, label %then, label %else
```

is false. Is this correctly handled in this implementation?

RolandMacDoland:
> Is this correctly handled in this implementation?

No. The case you describe above is an if/then/else clause I think. The fact is that these kinds of constructs are possible using `pyqir-generator` to mimic an if/then clause (ie there is this extra `else` block before joining to the same output block). Tbh, more sophisticated conditional constructs than if/then would require CFG analysis to satisfy SSA dominance principle.

cqc-alec (changes requested):
We need to support arbitrary control flow involving forward branching. For example, we need to support this kind of thing:

[recursion-3.pdf](https://github.com/CQCL/pytket-qir/files/10267383/recursion-3.pdf)

So a rethink is required. The "if-conversion" algorithm [1] could be useful.

[1] https://dl.acm.org/doi/abs/10.1145/567067.567085

RolandMacDoland:
> We need to support arbitrary control flow involving forward branching. For example, we need to support this kind of thing:
> 
> [recursion-3.pdf](https://github.com/CQCL/pytket-qir/files/10267383/recursion-3.pdf)
> 
> So a rethink is required. The "if-conversion" algorithm [1] could be useful.
> 
> [1] https://dl.acm.org/doi/abs/10.1145/567067.567085

Sure. However, I would be interested in running this example in the current implementation to see if it totally breaks it.

pytket_qir/gatesets/pyqir/pyqir.py
```diff
@@ -12,41 +12,67 @@
...
 _PYQIR_TO_TK[
-    QirGate(opnat=FuncNat.QIS, opname=FuncName.CNOT, opspec=FuncSpec.BODY)
+    QirGate(func_nat=FuncNat.QIS, func_name=FuncName.CNOT, func_spec=FuncSpec.BODY)
 ] = OpType.CX
 
 
 PYQIR_GATES = CustomGateSet(
     name="PyQir",
-    template=Template("__quantum__${opnat}__${opname}__${opspec}"),
+    template=Template("__quantum__${func_nat}__${func_name}__${func_spec}"),
```
cqc-melf, line 75: I think you mentioned a incompatibility with the format the function is written in some cases with L3? Why is that something you can't solve with updating this?

pytket_qir/generator.py
```diff
@@ -313,15 +313,17 @@ def condition_zero_block():
...
+                    func_spec=FuncSpec.BODY,
                 )
 
                 # Update gateset.
                 gateset = PYQIR_GATES.gateset
                 gateset["wasm"] = CustomQirGate(
-                    opnat=FuncNat.HYBRID,
-                    opname=WasmName.WASM,
-                    opspec=FuncSpec.BODY,
+                    func_nat=FuncNat.HYBRID,
```
cqc-melf, line 319: Is this update because of an external change? Or is that just a rename?

poetry.lock
```diff
@@ -75,7 +75,7 @@ test = ["codecov (>=2.1)", "pytest (>=7.2)", "pytest-cov (>=4.0)"]
 
 [[package]]
```
cqc-melf, line 76: Is this file automatically generated? Do we want to have it in the repo?

pytket_qir/converter.py
```diff
@@ -0,0 +1,357 @@
...
+        return topological_ordering
+    else:
+        raise ConverterException(
+            "A cycle exists in the graph, there is no topological order to be found."
+        )
+
+
+class QirConverter:
+    """
+    A class to build and convert the QIR control-flow graph to a pytket circuit.
```
cqc-melf, line 84: Maybe it would be good to have a few more word here how this works and what the steps in the execution are?

pytket_qir/converter.py
```diff
@@ -0,0 +1,357 @@
...
+        if optimisation_level == 1:
+            self.collapse_blocks()
+        self.circuit = self.cfg_to_circuit()
+
+    @property
+    def successors(self):
+        return self._successors
+
+    @successors.setter
+    def successors(self, value):
```
cqc-melf, line 188: Can you maybe add a comment for what this is used?

pytket_qir/converter.py
```diff
@@ -0,0 +1,357 @@
...
+                    or el == self._successors[block][1]
+                )
+                self._successors[block] = list(filter(f, self._successors))
+
+    @property
+    def predecessors(self):
+        return self._predecessors
+
+    @predecessors.setter
+    def predecessors(self, value):
```
cqc-melf, line 231: Can you maybe add a comment for what this is used?

pytket_qir/converter.py
```diff
@@ -0,0 +1,357 @@
...
+    @local_conditions.setter
+    def local_conditions(self, value):
+        self._local_conditions = value
+
+    @property
+    def cfg(self):
+        return self._cfg
+
+    @cfg.setter
+    def cfg(self, value) -> None:
```
cqc-melf, line 275: Can you maybe add a comment for what this is used?

pytket_qir/converter.py
```diff
@@ -0,0 +1,357 @@
...
+        # import pdb; pdb.set_trace()
```
cqc-melf: Is this import intended to stay here?

pytket_qir/converter.py
```diff
@@ -0,0 +1,357 @@
...
+        self.conditions = {}
+        self.edges = {}
+        self.local_conditions = {}
+        self.successors = {}
```
cqc-melf, line 287: This will call the function from above? can you maybe add a comment to point this out?

pytket_qir/converter.py
```diff
@@ -0,0 +1,357 @@
...
+                )
+                # f = lambda b: b if b.condition else None
+                self._cfg[block_name] = block_inst
+
+    def cfg_to_circuit(self) -> Circuit:
+        main_circuit = Circuit(self.parser.qubits, self.parser.bits + 2)
+        # Setting condition and local condition to True
+        # for the predecessor of the entry block.
+        main_circuit.add_c_setbits([1, 1], [self.parser.bits, self.parser.bits + 1])
+        condition_bit = main_circuit.get_c_register("c")[self.parser.bits]
```
cqc-melf, line 313: Are there any plans how to invert this bits when converting back to qir? Have you thought about putting special label on them?

pytket_qir/converter.py
```diff
@@ -0,0 +1,357 @@
...
+            circ_box = CircBox(circuit)
+            arguments = [qubit.index[0] for qubit in circuit.qubits] + [
+                bit.index[0] for bit in circuit.bits
+            ]  # Order matters.
+            main_circuit.add_circbox(
+                circ_box, arguments, condition=if_bit(self.conditions[block_name])
+            )
+        return main_circuit
+
+    def apply_contraction(self, block) -> Block:
```
cqc-melf: This is doing the fusion of two blocks? Or?

pytket_qir/converter.py
```diff
@@ -0,0 +1,357 @@
...
+                )
+            else:  # Return the current block.
+                return curr_block
+        return curr_block
+
+    def collapse_blocks(self) -> None:
+        """
+        Detect and collapse chains of linear blocks.
+
+        Return a rewritten CFG where all blocks have in-degree exactly one
```
cqc-melf: How are you taking care of the `in-degree exactly one` in the case where you are combining two branches to one?

pytket_qir/parser.py
```diff
@@ -104,31 +96,31 @@
...
-        self.module = QirModule(file_path)
+        self.module: QirModule = qir_module
         self.gateset: CustomGateSet = gateset if gateset is not None else PYQIR_GATES
         self.wasm_handler = wasm_handler
         self.wasm_int_type = wasm_int_type
         self.qir_int_type = qir_int_type
         self.qubits = self.get_required_qubits()
         self.bits = self.get_required_results()
-        self.set_cregs: Dict[str, BitRegister] = {}  # Keep track of set registers.
+        self.ssa_vars: Dict[str, BitRegister] = {}  # Log of set ssa variables.
```
cqc-melf, line 120: Are this the big registers for the conditions or is this something else?

tests/conftest.py
```diff
@@ -366,3 +1223,681 @@ def pytket_nested_conditionals_circuit(
...
+    }
+    return cfg
+
+
+@fixture
+def collapsed_nested_chains() -> dict:
+    cfg = {
+        "entry": Block(
+            name="entry",
+            succs=["else0", "then0"],
```
cqc-melf, line 1481: What are the conditions for the following nodes? Or is that not relevant in this data structure?

tests/qir_test_files/RUSLoopXX-1.ll
```diff
@@ -0,0 +1,131 @@
+; ModuleID = 'RUSLoopXX-1.bc'
```
cqc-melf, line 1: How have you generated this files?

cqc-melf: (Same question for the BC files)

cqc-melf (commented):
I have added a few questions and comments.

pytket_qir/gatesets/pyqir/pyqir.py
RolandMacDoland: Because it is a generic template for all function names. This is a convention from the QIR specs. Maybe a specialised version of this would work ? However, this would defeat the point in having a template though.

pytket_qir/generator.py
RolandMacDoland: It is a better renaming really. Previously, they were "operation" names but it is more accurate to talk about function names.

poetry.lock
RolandMacDoland: Yes, you version the lock file so that poetry knows exactly which dependency environment I have been developing this.

pytket_qir/converter.py
RolandMacDoland: Done.

pytket_qir/converter.py
RolandMacDoland: Done.

pytket_qir/converter.py
RolandMacDoland: Done.

pytket_qir/converter.py
RolandMacDoland: Done.

pytket_qir/converter.py
RolandMacDoland: Ah, sorry. I forgot to push the commit where I did some code clean up.

pytket_qir/converter.py
RolandMacDoland: Done.

pytket_qir/converter.py
RolandMacDoland: Actually, this is the responsibility of the `pytket_qir.generator`. I am currently thinking about how to do that.

pytket_qir/converter.py
RolandMacDoland: Yes exactly. I've added a comment.

pytket_qir/converter.py
RolandMacDoland: Good point, this is inexact. I've amended it.

pytket_qir/converter.py
```diff
@@ -0,0 +1,371 @@
...
+@dataclass
+class Block:
+    name: str
+    succs: List
+    preds: List
+    composition: List
+    visited: bool = False
+
+
+def topological_sort(digraph: dict) -> dict:
```
cqc-alec, line 52: Should we use a third-party library like `networkx` for graph manipulation? Then we'd be able to use the build-in algorithms, such as top sort and cyclicity checking.

pytket_qir/parser.py
RolandMacDoland: Not necessarily. These are maps from SSA variable names to the corresponding `pytket` register. It is not restricted to only conditions but any SSA variable (for instance as the result of some arithmetic operation). In the `QirConverter`, I filter this map to see if there is one with the right SSA variable name for the condition.

tests/conftest.py
RolandMacDoland: I am not sure this is relevant here. I added them (conditions and local conditions) to the `Block` type but since they are computed dynamically, I found it simpler to just use the class member dicts `self.conditions` and `self.local_conditions`.

tests/qir_test_files/RUSLoopXX-1.ll
RolandMacDoland: No. This one is an example that has been provided to me by Pete.

pytket_qir/converter.py
RolandMacDoland: Could be. One thing that can be interesting is if `networkx` allows for a user to define custom graph rewrites more easily. Then, one can come up with other and better CFG optimisations (than my naive one about block contractions) and apply them. At least, it could provide a better framework for doing this. However, after a quick glance at the doc, it doesn't seem to provide that functionality. Later, in case of QIR programs that exhibit back edges, that can be interesting if loops aren't fully unrolled.

pytket_qir/converter.py
cqc-melf: It could be helpful to mark this somehow to make it easier to detect them later?

RolandMacDoland: Hm, not sure I follow you. To mark what exactly ?

pytket_qir/converter.py
cqc-melf: I have thought about replacing this with something else as well, but I thought that it might be better in the long term to not have that much optimisation going on in the mainly conversion module. I think we should try to parse this in the long term into something that allows us to optimise this on the c++ level?

RolandMacDoland: Could be. Although it could also fit into the TKET2 AST Seyon mentioned elsewhere. As far as I've understood, TKET2 is designed to handle conditionals and classical arithmetic in a better way than what TKET1 currently does. I am not sure how much the AST overlaps/captures the CFG structure but if it does to some extend, it would make sense that optimisations happen there.

pytket_qir/converter.py
cqc-melf: To mark that this bit as been created to store the block condition?

pytket_qir/converter.py
```diff
@@ -0,0 +1,368 @@
...
+                        del self.edges[block]
+                else:
+                    rewritten_cfg[block_name] = block
+        # Reset visited fields to false for later reuse.
+        for block in rewritten_cfg.values():
+            block.visited = False
+        self.rewritten_cfg = rewritten_cfg
+
+
+def circuit_from_qir(
```
cqc-melf, line 601: Here is still some documentation missing. Maybe it would be good to write a few words for each parameter to point out what they are used for?

pytket_qir/converter.py
```diff
@@ -359,6 +359,30 @@ def cfg_to_circuit(self) -> Circuit:
             )
         return main_circuit
 
+    def circuit_to_module(self):
+        self.generator.circuit_to_module(self.generator.circuit, self.generator.module)
+        return self.generator.module
+        # for command in self.circuit:
```
cqc-melf: Do you want to remove this?

pytket_qir/converter.py
```diff
@@ -475,10 +475,87 @@ def circuit_from_qir(
...
+    circuit = converter.to_circuit()
     # Attach few fields to the circuit.
     circuit.cfg = converter.rewritten_cfg
+    circuit.edges = converter.edges
     circuit.ssa_vars = converter.parser.ssa_vars
     return circuit
+
+
+def circuit_to_qir(
+    circuit: Circuit,
```
cqc-melf, line 653: Are there tests of this function which checks the generates QIR?

tests/converter_test.py
```diff
@@ -452,3 +452,15 @@ def test_collapse_nested_chains(self, collapsed_nested_chains) -> None:
         circuit = circuit_from_qir(collapse_nested_chains_bc, optimisation_level=1)
 
         assert circuit.cfg == collapsed_nested_chains
+
+
+class TestRoundTripForGuardedCircuits:
+    @pytest.mark.skip
```
cqc-melf: Is it intended that this is still skipped?

pytket_qir/generator.py
```diff
@@ -410,70 +406,3 @@ def condition_zero_block():
                     else:
                         get_gate(*qubits)
         return module
-
-
-def circuit_to_qir(
```
cqc-melf, line 413: Is it intended that this is removed again?

pytket_qir/converter.py
```diff
@@ -103,12 +103,13 @@ def __init__(
         wasm_int_size: int = 32,
         qir_int_size: int = 64,
         circuit: Optional[Circuit] = None,
-        module: Optional[SimpleModule] = None,
+        module: Optional[Module] = None,
```
cqc-melf, line 106: Is this just a rename?

pytket_qir/converter.py
```diff
@@ -126,14 +126,14 @@ def __init__(
                 self.collapse_blocks()
         if circuit is not None:
             self.circuit = circuit
-            self.module = module
+            self.module = cast(Module, module)
```
cqc-melf, line 130: Is this something mypy is requesting?

pytket_qir/converter.py
```diff
@@ -421,6 +398,69 @@ def _set_bit(self, exp, module, set_bits):
             )
             x = module.builder.call(source, [])
         return module, x
+    def _parse_logic_exp(self, exp: LogicExp, module: Module, set_bits: dict):
```
cqc-melf, line 403: LogicExp is from pytket?

pytket_qir/generator.py
```diff
@@ -406,3 +405,107 @@ def condition_zero_block():
                     else:
                         get_gate(*qubits)
         return module
+
+    def command_to_module(
+        self, command: Command, circuit: Circuit, module: Module
```
cqc-melf, line 410: what is the module parameter used for?

tests/converter_test.py
```diff
@@ -455,12 +466,44 @@ def test_collapse_nested_chains(self, collapsed_nested_chains) -> None:
...
+        qir_bytes = cast(bytes, circuit_to_qir(circuit, gateset=ext_pyqir_gates))
+
+        ll = str(bitcode_to_ir(qir_bytes))
+
+        import pdb
+
+        pdb.set_trace()
 
-        qir = circuit_to_qir(circuit1)
+        print("YAY")
```
cqc-melf: :)

pytket_qir/converter.py
RolandMacDoland: No, this is my custom type to extend `pyqir` module functionality to arbitrary gateset.

pytket_qir/converter.py
RolandMacDoland: Yes.

pytket_qir/converter.py
RolandMacDoland: Yes.

pytket_qir/generator.py
RolandMacDoland: It is populated with the corresponding LLVM instructions for the `pytket` circuit. Then, it can generate QIR by calling the `.ir()` or `.bitcode()` methods.

tests/converter_test.py
RolandMacDoland: Ahah, sorry about that. It is only to stop the debugger at that point.

tests/converter_test.py
```diff
@@ -465,6 +468,21 @@ def test_collapse_nested_chains(self, collapsed_nested_chains) -> None:
 
 
 class TestRoundTripForGuardedCircuits:
+
+    @pytest.mark.skip(reason="Test parsing the logical expression.")
```
cqc-melf: Is this intended to be skipped?

tests/qir_test_files/RoundTripDiamondConditional.ll
```diff
@@ -35,9 +35,8 @@ else:                                             ; preds = %entry
 
 continue:                                         ; preds = %else, %then
   %1 = call i1 @source()
-  %2 = and i1 true, %1
-  %3 = or i1 %2, true
-  br i1 %3, label %then1, label %else2
+  %2 = or i1 true, %1
```
cqc-melf, line 38: Why is %3 removed here?

pytket_qir/converter.py
```diff
@@ -410,64 +410,68 @@ def _parse_logic_exp(self, exp: LogicExp, module: Module, set_bits: dict):
...
                     self.ssa_vars["%last"] = ssa_var
+                    return module, ssa_var
+                else:
+                    module, ssa_var1 = self._parse_logic_exp(bitor_arg, module, set_bits)
+                    module, ssa_var2 = self._parse_logic_exp(bitand_arg, module, set_bits)
+                    _, build = _TK_TO_PYQIR_LOGIC[bitwiseop]
+                    ssa_var = build(module.builder)(ssa_var1, ssa_var2)
                     self.ssa_vars["%last"] = ssa_var
+                    return module, ssa_var
             else:
```
cqc-melf, line 476: I understand this refactor in a way that it only moves out the negation from the else case?

tests/converter_test.py
RolandMacDoland: It was a quick and dirty check but I made it a proper test in the last commits.

pytket_qir/converter.py
RolandMacDoland: Yes. I wanted it to be more in line with the recursive nature of that function. In essence, I wanted to recurse also when it detected a negated bit because this is a subtype of `LogicExp`.

cqc-melf (approved):

PR #27, opened 2023-05-11
first steps for conversion and handling of simple conditionals

cqc-melf:
add simple conversion of pytket circuits into qir
this can handle:
* classical operations
* simple conditionals
* quantum gates
* measurements into default register


pytket/qir/conversion/module.py
```diff
@@ -0,0 +1,48 @@
...
+from typing import Optional
+
+from pyqir import SimpleModule, BasicQisBuilder  # type: ignore
+
+from pytket.wasm import WasmFileHandler
+
+from .gatesets import PYQIR_GATES
+
+
+class Module:
```
cqc-alec: This is a rather overloaded name; how about something like `PytketModule`?

pytket/qir/conversion/api.py
```diff
@@ -0,0 +1,50 @@
...
+
+from pytket.circuit import Circuit
+
+from .conversion import QirGenerator
+from .module import Module
+
+
+def pytket_to_qir(circ: Circuit) -> str:
+    """converts give circ to qir string"""
+    return str(circ)
```
cqc-alec: I guess this is just a dummy function at the moment?

pytket/qir/conversion/api.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def pytket_to_qir_2(circ: Circuit) -> str:
```
cqc-alec: Why two functions?

pytket/qir/conversion/conversion.py
```diff
@@ -0,0 +1,360 @@
...
+}
+
+
+class QirGenerator:
+    """Generate QIR from a pytket circuit."""
+
+    def __init__(
+        self,
+        circuit: Circuit,
+        module: Module,
```
cqc-alec: Since this class is just for generating a `Module` from a `Circuit`, why does the user need to initialize it with a `Module`? The `circuit_to_module()` method takes _another_ module as an argument, which is quite confusing too.

(If this is intended as an "internal" class used by the `pytket_to_qir()` functions(s) then I guess it's OK (I see that the methods are called recursively), but still needs documenting as to how it should be used.)

pytket/qir/conversion/conversion.py
```diff
@@ -0,0 +1,360 @@
...
+            elif op.get_name() == "XOR":
+                optype = BitWiseOp.XOR
+        elif optype in [OpType.Barrier, OpType.CopyBits]:
+            pass
+        else:
+            params = op.params
+        return (optype, params)
+
+    def _to_qis_qubits(self, qubits: List[Qubit]) -> Sequence[Qubit]:
+        return [self.module.module.qubits[qubit.index[0]] for qubit in qubits]
```
cqc-alec, line 163: This will assign the same QIR qubit to qubits from different registers having the same index. We should probably instead use the index of `qubit` in `self.circuit.qubits`, to ensure a 1-1 map.

Or are we requiring the circuit to have a single register?

pytket/qir/conversion/conversion.py
```diff
@@ -0,0 +1,360 @@
...
+
+    def _to_qis_results(self, bits: List[Bit]) -> Optional[Value]:
+        if bits:
+            return self.module.module.results[bits[0].index[0]]  # type: ignore
+        return None
+
+    def _to_qis_bits(self, args: List[Bit]) -> Sequence[Value]:
+        if args:
+            return [self.module.module.results[bit.index[0]] for bit in args[:-1]]
```
cqc-alec, line 174: Same comment for bits.

pytket/qir/conversion/conversion.py
```diff
@@ -0,0 +1,360 @@
...
+        """Populate a PyQir module from a pytket circuit."""
+        for command in circuit:
+            op = command.op
+            if isinstance(op, Conditional):
+                assert op.width == 1  # only one conditional bit
+
+                conditional_circuit = self._rebase_op_to_gateset(
+                    op.op, command.args[op.width :]
+                )
+                condition_bit_index = command.args[0].index[0]
```
cqc-alec, line 285: See other comments regarding use of `.index[0]`.

pytket/qir/conversion/conversion.py
```diff
@@ -0,0 +1,360 @@
...
+
+                # condition_ssa = module.module.results[condition_bit_index]
+
+                def condition_block() -> None:
+                    """
+                    Populate recursively the module with the contents of the conditional
+                    sub-circuit when the condition is True.
+                    """
+                    if op.value == 1:
+                        self.circuit_to_module(conditional_circuit, module)
```
cqc-alec, line 296: This confuses me.

* `op.value` is the bit value according to which the circuit should be executed at runtime, right? What's the meaning of these two lines?
* What's with the `true = lambda: condition_block()` lines below, when `condition_block()` is a function that returns a `Module`?
* What happens for other values of `op.value`?

pytket/qir/conversion/conversion.py
```diff
@@ -0,0 +1,360 @@
...
+
+                conditional_circuit = self._rebase_op_to_gateset(
+                    op.op, command.args[op.width :]
+                )
+                condition_bit_index = command.args[0].index[0]
+                condition_name = command.args[0].reg_name
+
+                # condition_ssa = module.module.results[condition_bit_index]
+
+                def condition_block() -> None:
```
cqc-alec, line 290: Looks like this returns a `Module`, not `None`?

pytket/qir/conversion/conversion.py
cqc-melf: We are requiring the circuit to have a single register, qir can only handle one register so there is no way to handle more than one.

pytket/qir/conversion/conversion.py
cqc-melf: This function should only be used for in the standard c register, all other bits can't be converted to converted to the result type. I will add an assertion for this.

pytket/qir/conversion/conversion.py
cqc-melf: This should be the index of the bit in its register, I think this should be the right way?

pytket/qir/conversion/conversion.py
cqc-melf: It has no return statement, the return value from circuit_to_module is not stored?

pytket/qir/conversion/api.py
cqc-melf: yes

cqc-melf:
I would merge this now as it is and we can continue the discussion later @cqc-alec ?

cqc-alec (approved):

PR #35, opened 2023-05-24
fix entry point handling

cqc-melf:


cqc-alec:
When using this conversion function we get a different error back:

```
Expected the QIR file to have qubit count specifiedbut no annotation was found
```

So it looks like it's solved one problem but there is a similar problem with the `num_required_qubits` attribute.

cqc-melf:
> When using this conversion function we get a different error back:
> 
> ```
> Expected the QIR file to have qubit count specifiedbut no annotation was found
> ```
> 
> So it looks like it's solved one problem but there is a similar problem with the `num_required_qubits` attribute.

I have updated that as well

pytket/qir/conversion/api.py
```diff
@@ -83,3 +85,33 @@ def pytket_to_qir(
...
+    :type name: str
+    :param returntype: format of the generated qir, defaut value is binary
+    :type returntype: ReturnTypeQIR
+    """
+
+    initial_result = str(pytket_to_qir(circ, name, ReturnTypeQIR.STRING))  # type: ignore
+
+    initial_result = initial_result.replace("entry_point", "EntryPoint")  # type: ignore
+    initial_result = initial_result.replace("num_required_qubits", "requiredQubits")  # type: ignore
+    result = initial_result.replace("num_required_results", "requiredResults")  # type: ignore
```
cqc-alec: I tried this locally and now the submission fails with a completely unhelpful error message:

```
E               pytket.extensions.quantinuum.backends.quantinuum.GetResultFailed: Cannot retrieve result; job status is CircuitStatus(status=<StatusEnum.ERROR: 'Circuit has errored. Check CircuitStatus.message for error message.'>, message='{"name": "job", "submit-date": "2023-05-24T15:40:02.098446", "result-date": null, "queue-position": null, "cost": null, "error": null}', error_detail=None, completed_time=None, queued_time=None, submitted_time=None, running_time=None, cancelled_time=None, error_time=None, queue_position=None)
```

cqc-melf: Interesting, is that what we are expecting because of the missing register handling?

cqc-alec: I guess we would expect it to fail, yeah.

```suggestion
    result = initial_result.replace("entry_point", "EntryPoint").replace("num_required_qubits", "requiredQubits").replace("num_required_results", "requiredResults")  # type: ignore
```

pytket/qir/conversion/api.py
```diff
@@ -83,3 +85,33 @@ def pytket_to_qir(
...
+
+    bitcode = pyqir.Module.from_ir(pyqir.Context(), result).bitcode  # type: ignore
```
cqc-alec: No need to compute this unless it's required.

pytket/qir/conversion/api.py
```diff
@@ -83,3 +85,33 @@ def pytket_to_qir(
         return populated_module.module.ir()
     else:
         raise ValueError("unsupported return type")
+
+
+def pytket_to_qir_optimised(
```
cqc-alec: Seems a strange name?

If we're going to merge this I think we should keep the same function name as before rather than add a new one. We could give it an optional argument like `pyqir_0_7_compatibility` defaulting to `False`, so that this extra code is only run if requested.

cqc-alec (changes requested):

pytket/qir/conversion/api.py
cqc-melf: I wanted to have this in all cases to make sure that the replacements that we are doing still generated results that can be read by pyqir.

pytket/qir/conversion/api.py
cqc-melf: done in @cqc-melf
[update function name and add parameter for compability](https://github.com/CQCL/pytket-qir/pull/35/commits/f7d15a8b521ba9eb9366304988983b5b121ab8a9)
[f7d15a8](https://github.com/CQCL/pytket-qir/pull/35/commits/f7d15a8b521ba9eb9366304988983b5b121ab8a9)

pytket/qir/conversion/api.py
```diff
@@ -38,6 +40,7 @@ def pytket_to_qir(
     circ: Circuit,
     name: str = "Generated from input pytket circuit",
     returntype: ReturnTypeQIR = ReturnTypeQIR.BINARY,
+    pyqir_0_7_compatibility: bool = False,
```
cqc-alec, line 43: Should document the effect of this in the docstring below.

pytket/qir/conversion/api.py
```diff
@@ -77,9 +80,26 @@ def pytket_to_qir(
...
+        result = initial_result.replace("entry_point", "EntryPoint").replace("num_required_qubits", "requiredQubits").replace("num_required_results", "requiredResults")  # type: ignore
+
+        bitcode = pyqir.Module.from_ir(pyqir.Context(), result).bitcode  # type: ignore
+
+        if returntype == ReturnTypeQIR.BINARY:
+            return bitcode  # type: ignore
+        elif returntype == ReturnTypeQIR.STRING:
+            return result  # type: ignore
+        else:
+            raise ValueError("unsupported return type")
```
cqc-alec: I think we can actually have an `assert` here, since we know these are the only values of the enum.

tests/api_test.py
```diff
@@ -80,6 +85,59 @@ def test_pytket_qir() -> None:
...
+    circ.H(0)
+
+    result = pytket_to_qir(
+        circ,
+        name="test_pytket_qir",
+        returntype=ReturnTypeQIR.STRING,
+        pyqir_0_7_compatibility=True,
+    )
+
+    print(result)
```
cqc-alec, line 95:
```suggestion
```

tests/api_test.py
```diff
@@ -109,6 +167,62 @@ def test_pytket_api_creg() -> None:
...
+    circ = Circuit(3)
+    circ.H(0)
+
+    result = pytket_to_qir(
+        circ, name="test_pytket_qir", returntype=ReturnTypeQIR.STRING
+    )
+
+    result_2 = pytket_to_qir(circ, name="test_pytket_qir")
+
+    print(result)
```
cqc-alec, line 174:
```suggestion
```

pytket/qir/conversion/api.py
```diff
@@ -77,9 +80,26 @@ def pytket_to_qir(
...
-    if returntype == ReturnTypeQIR.BINARY:
-        return populated_module.module.bitcode()
-    elif returntype == ReturnTypeQIR.STRING:
-        return populated_module.module.ir()
+
+    if pyqir_0_7_compatibility:
+
+        initial_result = str(populated_module.module.ir())  # type: ignore
+
+        result = initial_result.replace("entry_point", "EntryPoint").replace("num_required_qubits", "requiredQubits").replace("num_required_results", "requiredResults")  # type: ignore
```
cqc-alec, line 89: I guess we don't have black checks on the CI. Would be good to add (in another PR).

cqc-alec (changes requested):

pytket/qir/conversion/api.py
cqc-melf: we have that, somehow this is fine with black

tests/api_test.py
cqc-melf: Do we want to remove all the prints of the result? They are currently in all of the testcases and I think it is somehow helpful to have them at the moment in case of any more changes?

pytket/qir/conversion/api.py
cqc-alec: Surprising, but OK!

tests/api_test.py
cqc-alec: At some point we should, but if they're helpful at the moment we can keep them.

pytket/qir/conversion/api.py
cqc-melf: done @cqc-melf
[replace valuerror for qir_format qith assertion](https://github.com/CQCL/pytket-qir/pull/35/commits/bbed6f40ababe52a4e1e8ef1176f1abe4dceeb71)
[bbed6f4](https://github.com/CQCL/pytket-qir/pull/35/commits/bbed6f40ababe52a4e1e8ef1176f1abe4dceeb71)

pytket/qir/conversion/api.py
cqc-melf: Done in the merge commit

cqc-alec (approved):

PR #37, opened 2023-05-24
Update/bigger conditional

cqc-melf:


pytket/qir/conversion/conversion.py
```diff
@@ -493,50 +493,98 @@ def circuit_to_module(
...
+                            ssabool,
+                            true=lambda: condition_block_true(),  # type: ignore
+                            false=lambda: condition_block_false(),  # type: ignore
+                        )
 
-                if condition_name in self.ssa_vars:
+                    else:
+                        ValueError(
+                            """circuit contruction with this condition
+                            not yet supported"""
```
cqc-alec: What does this error message mean? Can it explain exactly what is not yet supported?

pytket/qir/conversion/conversion.py
```diff
@@ -493,50 +493,98 @@ def circuit_to_module(
...
+                        )
+                        # this should be an assertion when working
```
cqc-alec: Not sure what this comment means.

pytket/qir/conversion/conversion.py
```diff
@@ -493,50 +493,98 @@ def circuit_to_module(
...
+                else:
+                    conditional_circuit = self._rebase_op_to_gateset(
```
cqc-alec: There seems to be some repeated code here; can it be factored out to before the line `if op.width == 1`?

pytket/qir/conversion/conversion.py
```diff
@@ -493,50 +493,98 @@ def circuit_to_module(
...
-                            pyqir.const(self.qir_int_type, condition_bit_index),
-                        ],
+                    for i in range(op.width):
+                        if command.args[i].reg_name != condition_name:
+                            raise ValueError(
+                                "conditional can only work with one entire register"
+                            )
+
+                    for i in range(op.width - 1):
+                        if command.args[i].index[0] >= command.args[i + 1].index[0]:
```
cqc-alec, line 546: Do we need to insist that the bit indices are monotonically increasing? We should be able to handle e.g. `condition_bits=[1,0], condition_value=2` which is equivalent to `condition_bits=[0,1], condition_value=1` (just needs some bit twiddiling on the value).

pytket/qir/conversion/conversion.py
```diff
@@ -493,50 +493,98 @@ def circuit_to_module(
...
+                    lower_cond = module.module.builder.icmp(
+                        pyqir.IntPredicate.SGT,
+                        pyqir.const(self.qir_int_type, op.value),
+                        self.ssa_vars[condition_name],
+                    )
+                    upper_cond = module.module.builder.icmp(
+                        pyqir.IntPredicate.SGT,
+                        self.ssa_vars[condition_name],
+                        pyqir.const(self.qir_int_type, op.value),
                     )
```
cqc-alec: Do we really need to test for a lower bound and an upper bound when they are the same? Can't we just test for equality?

cqc-alec (changes requested):

pytket/qir/conversion/conversion.py
cqc-melf: This means that the error should be removed when all conditionals are supported. I would hope that this is now the case. I have removed this and added an assertion to replace it.

pytket/qir/conversion/conversion.py
cqc-melf: I have clean this up.

pytket/qir/conversion/conversion.py
cqc-melf: I have copied this checks from the qasm conversion. Do we want to support things that have not been possible in qasm? It would be possible as well to do subsets of a register as well if we want to? I have just skipped that for the same reason.

pytket/qir/conversion/conversion.py
cqc-melf: I have combined that to one check, see @cqc-melf
[clean up conditional handling](https://github.com/CQCL/pytket-qir/pull/37/commits/43eefb9ba2e33b72efc8026785a99f7596aaf385)
[43eefb9](https://github.com/CQCL/pytket-qir/pull/37/commits/43eefb9ba2e33b72efc8026785a99f7596aaf385)

pytket/qir/conversion/conversion.py
cqc-melf: See @cqc-melf
[clean up conditional handling](https://github.com/CQCL/pytket-qir/pull/37/commits/43eefb9ba2e33b72efc8026785a99f7596aaf385)
[43eefb9](https://github.com/CQCL/pytket-qir/pull/37/commits/43eefb9ba2e33b72efc8026785a99f7596aaf385)

pytket/qir/conversion/conversion.py
cqc-melf: See @cqc-melf
[clean up conditional handling](https://github.com/CQCL/pytket-qir/pull/37/commits/43eefb9ba2e33b72efc8026785a99f7596aaf385)
[43eefb9](https://github.com/CQCL/pytket-qir/pull/37/commits/43eefb9ba2e33b72efc8026785a99f7596aaf385)

pytket/qir/conversion/conversion.py
cqc-melf: See below, removed in @cqc-melf
[clean up conditional handling](https://github.com/CQCL/pytket-qir/pull/37/commits/43eefb9ba2e33b72efc8026785a99f7596aaf385)
[43eefb9](https://github.com/CQCL/pytket-qir/pull/37/commits/43eefb9ba2e33b72efc8026785a99f7596aaf385)

pytket/qir/conversion/conversion.py
cqc-alec: OK, if this is a restriction of the QASM converter it's probably fine for now.

cqc-alec (approved):

PR #38, opened 2023-05-25
Feature/add results recording

cqc-melf:


cqc-melf:
This should solve the problem with the missing names of the registers in the recording of the i64.

cqc-melf:
We could think about excluding some register from being recorded, like `_TKET... `

cqc-melf:
I think we should merge this, it looks like this at least solving some of the problems? @cqc-alec

pytket/qir/conversion/conversion.py
```diff
@@ -178,14 +178,54 @@ def __init__(
...
+                pyqir.FunctionType(
+                    pyqir.Type.void(self.module.module.context),
+                    [
+                        pyqir.IntType(self.module.module.context, qir_int_type),
+                        self.reg_const[list(self.reg_const)[0]].type,
+                    ],
+                ),
+            )
+
+            # void __quantum__rt__result_record_output(result)
```
cqc-alec: Delete commented-out code.

pytket/qir/conversion/conversion.py
```diff
@@ -747,13 +787,35 @@ def condition_block() -> None:
...
+                    reg_name = creg[0].reg_name
+                    self.module.builder.call(
+                        self.record_output_i64,
+                        [
+                            self.ssa_vars[reg_name],
+                            self.reg_const[reg_name],
+                        ],
+                    )
+
+                # for i in range(self.circuit.q_registers[0].size):
```
cqc-alec: Delete commented-out code.

pytket/qir/conversion/conversion.py
```diff
@@ -178,14 +178,54 @@ def __init__(
...
+        if len(self.reg_const) > 0:
+
+            # void __quantum__rt__int_record_output(i64)
+            self.record_output_i64 = self.module.module.add_external_function(
+                "__quantum__rt__int_record_output",
+                pyqir.FunctionType(
+                    pyqir.Type.void(self.module.module.context),
+                    [
+                        pyqir.IntType(self.module.module.context, qir_int_type),
+                        self.reg_const[list(self.reg_const)[0]].type,
```
cqc-alec: What's special about the zeroth member of `list(self.reg_const)`? What if there are multiple?

pytket/qir/conversion/conversion.py
```diff
@@ -747,13 +787,35 @@ def condition_block() -> None:
                     else:
                         get_gate(*qubits)
         if record_output:
-            for creg in self.circuit.c_registers:
-                reg_name = creg[0].reg_name
+
+            if len(self.reg_const) > 0:
```
cqc-alec: I don't think this condition should be necessary.

cqc-alec (changes requested):

pytket/qir/conversion/conversion.py
cqc-melf: From my understanding there is nothing special about the first element, they all should have the same type. The i8 pointer type is not directly available so I have to extract it from the given value.

pytket/qir/conversion/conversion.py
cqc-melf: It is required because we cant define the function `__quantum__rt__int_record_output` without the type of the constant.

pytket/qir/conversion/conversion.py
cqc-melf: done in @cqc-melf
[clean up](https://github.com/CQCL/pytket-qir/pull/38/commits/0c16847c2eb3ff7a2d33a6203c82b39be8c28564)
[0c16847](https://github.com/CQCL/pytket-qir/pull/38/commits/0c16847c2eb3ff7a2d33a6203c82b39be8c28564)

pytket/qir/conversion/conversion.py
cqc-melf: done in @cqc-melf
[clean up](https://github.com/CQCL/pytket-qir/pull/38/commits/0c16847c2eb3ff7a2d33a6203c82b39be8c28564)
[0c16847](https://github.com/CQCL/pytket-qir/pull/38/commits/0c16847c2eb3ff7a2d33a6203c82b39be8c28564)

pytket/qir/conversion/conversion.py
cqc-melf: We could add an additional constant to avoid this if you want to?

pytket/qir/conversion/conversion.py
cqc-alec: Can we do something like this?
https://github.com/qir-alliance/pyqir/blob/43de8247c297e47dd5602d39ba7f77471ccd87b6/pyqir/tests/test_rt.py#L16

cqc-melf: Yes, good point. Done in @cqc-melf
[add point type](https://github.com/CQCL/pytket-qir/pull/38/commits/3439488874d80935be430393f16f0ac0aae1dc4a)
[3439488](https://github.com/CQCL/pytket-qir/pull/38/commits/3439488874d80935be430393f16f0ac0aae1dc4a)

pytket/qir/conversion/conversion.py
cqc-melf: remove in @cqc-melf
[add point type](https://github.com/CQCL/pytket-qir/pull/38/commits/3439488874d80935be430393f16f0ac0aae1dc4a)
[3439488](https://github.com/CQCL/pytket-qir/pull/38/commits/3439488874d80935be430393f16f0ac0aae1dc4a)

cqc-alec (approved):

PR #40, opened 2023-05-30
try to add a working qir example

cqc-melf:


pytket/qir/conversion/api.py
```diff
@@ -85,9 +85,34 @@ def pytket_to_qir(
 
     if pyqir_0_7_compatibility:
 
+        assert len(circ.c_registers) <= 1
```
cqc-alec: Is this restriction necessary? Even if it is we should presumably raise an error rather than assert.

pytket/qir/conversion/api.py
```diff
@@ -85,9 +85,34 @@ def pytket_to_qir(
...
+            elif "@set_all_bits_in_reg" in line:
+                keep_line = False
+
+            if keep_line:
+                result += line
+                result += "\n"
+
+        result = result.replace("i64 %0", "i64 0")
+        for _ in range(10):
+            result = result.replace("\n\n\n\n", "\n\n")
```
cqc-alec: Can we do this in a more functional way, something like

```python
def keep_line(line):
    return "@__quantum__qis__read_result__body" not in line and ...

result = "\n".join(filter(keep_line, initial_result.split("\n")))
```

cqc-alec (changes requested):

pytket/qir/conversion/api.py
cqc-melf: Yes it is at the moment. If we want to we could add more complicated code here to make sure this is handled in the right way for more registers. I have changes this to raise an error.

pytket/qir/conversion/api.py
cqc-melf: done in @cqc-melf
[requested changes](https://github.com/CQCL/pytket-qir/pull/40/commits/58161b6be18b8a83a78ffd6981bdf9e0fcc00a46)
[58161b6](https://github.com/CQCL/pytket-qir/pull/40/commits/58161b6be18b8a83a78ffd6981bdf9e0fcc00a46)

pytket/qir/conversion/api.py
cqc-melf: done in @cqc-melf
[requested changes](https://github.com/CQCL/pytket-qir/pull/40/commits/58161b6be18b8a83a78ffd6981bdf9e0fcc00a46)
[58161b6](https://github.com/CQCL/pytket-qir/pull/40/commits/58161b6be18b8a83a78ffd6981bdf9e0fcc00a46)

pytket/qir/conversion/api.py
```diff
@@ -85,9 +85,30 @@ def pytket_to_qir(
...
+                ("@__quantum__qis__read_result__body" not in line)
+                and ("@set_one_bit_in_reg" not in line)
+                and ("@reg2var" not in line)
+                and ("@read_bit_from_reg" not in line)
+                and ("@set_all_bits_in_reg" not in line)
+            )
+
+        result = "\n".join(filter(keep_line, initial_result.split("\n")))
+
+        result = result.replace("i64 %0", "i64 0")
```
cqc-alec, line 110: What is the effect of this? Probably needs a comment at least?

pytket/qir/conversion/api.py
```diff
@@ -85,9 +85,30 @@ def pytket_to_qir(
...
+        for _ in range(10):
+            result = result.replace("\n\n\n\n", "\n\n")
```
cqc-alec, line 114: Where have all the extra `"\n"` come from?

pytket/qir/conversion/api.py
cqc-melf: It will replace the ssa variable of the register which we are not creating any more with an integer it can use (in this case 0)

pytket/qir/conversion/api.py
cqc-melf: It removes the empty lines between the function declarations which add up. We can't remove the lines together with the functions because the empty lines only exists between the functions and not all functions get added for all the given circuits.

cqc-melf:
Keep in mind that all of this should be removed again after the backend is able to handle all the functions we are using for the register handling.

cqc-alec (approved):

PR #52, opened 2023-07-20
add a special case handling for eq predicate

cqc-melf:
This PR is adding a special case handling for all Rangepredicates were the two arguments are identical.
When the lower and the upper bound of the predicate are identical the op should be translated to one `pyqir.IntPredicate.EQ` and not to three instructions like it is done for all other cases.

### DIFF
--- a/pytket/qir/conversion/conversion.py
+++ b/pytket/qir/conversion/conversion.py
@@ -491,31 +491,55 @@ class QirGenerator:
             op = command.op
 
             if isinstance(op, RangePredicateOp):
-                lower_qir = pyqir.const(self.qir_int_type, op.lower)
-                upper_qir = pyqir.const(self.qir_int_type, op.upper)
+                # special case handling for REG_EQ
+                if op.lower == op.upper:
+                    registername = command.args[0].reg_name
 
-                registername = command.args[0].reg_name
-
-                lower_cond = module.module.builder.icmp(
-                    pyqir.IntPredicate.SGT, lower_qir, self.ssa_vars[registername]
-                )
-                upper_cond = module.module.builder.icmp(
-                    pyqir.IntPredicate.SGT, self.ssa_vars[registername], upper_qir
-                )
-
-                result = module.module.builder.and_(lower_cond, upper_cond)
-
-                condition_bit_index = command.args[-1].index[0]
-                registername = command.args[-1].reg_name
-
-                self.module.builder.call(
-                    self.set_one_bit_in_reg,
-                    [
+                    result = module.module.builder.icmp(
+                        pyqir.IntPredicate.EQ,
+                        pyqir.const(self.qir_int_type, op.lower),
                         self.ssa_vars[registername],
-                        pyqir.const(self.qir_int_type, condition_bit_index),
-                        result,
-                    ],
-                )
+                    )
+
+                    condition_bit_index = command.args[-1].index[0]
+                    result_registername = command.args[-1].reg_name
+
+                    self.module.builder.call(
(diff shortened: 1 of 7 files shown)


pytket/qir/conversion/conversion.py
```diff
@@ -491,31 +491,55 @@ def circuit_to_module(
...
+                        self.ssa_vars[registername],
+                    )
 
-                lower_cond = module.module.builder.icmp(
-                    pyqir.IntPredicate.SGT, lower_qir, self.ssa_vars[registername]
-                )
-                upper_cond = module.module.builder.icmp(
-                    pyqir.IntPredicate.SGT, self.ssa_vars[registername], upper_qir
-                )
+                    condition_bit_index = command.args[-1].index[0]
```
CalMacCQ, line 504: Why is the condition bit related to the last arg in the command?

CalMacCQ (commented):
As far as I can tell this looks fine. However I have no experinece with QIR.

Could you please update the PR description explaining briefly the purpose of this change?

pytket/qir/conversion/conversion.py
cqc-melf: That is just how it is defined in pytket

pytket/qir/conversion/conversion.py
```diff
@@ -491,31 +491,55 @@ def circuit_to_module(
...
+                        [
+                            self.ssa_vars[result_registername],
+                            pyqir.const(self.qir_int_type, condition_bit_index),
+                            result,
+                        ],
+                    )
 
-                condition_bit_index = command.args[-1].index[0]
-                registername = command.args[-1].reg_name
+                else:
```
cqc-melf, line 516: Github messed the commit up a bit showing what is old and what is new. All of old part got moved here in the else area and the part in the if case is new.

tests/qir/test_pytket_qir_14.ll
```diff
@@ -72,28 +72,24 @@ entry:
   call void @set_all_bits_in_reg(i64 %0, i64 %17)
   %18 = lshr i64 %0, 1
   call void @set_all_bits_in_reg(i64 %1, i64 %18)
-  %19 = icmp sgt i64 1, %0
```
cqc-melf, line 75: This file is part of the tests, it is just used to compared the generated qir to what we are expecting

tests/qir/test_pytket_qir_14.ll
```diff
@@ -102,30 +98,24 @@ else:                                             ; preds = %entry
   br label %continue
 
 continue:                                         ; preds = %else, %then
-  %35 = call i1 @read_bit_from_reg(i64 %0, i64 0)
```
cqc-melf, line 105: This file is part of the tests, it is just used to compared the generated qir to what we are expecting

tests/qir/test_pytket_qir_conditional_7.ll
```diff
@@ -0,0 +1,124 @@
+; ModuleID = 'test_pytket_qir_conditional_7'
```
cqc-melf, line 1: This file is part of the tests, it is just used to compared the generated qir to what we are expecting

tests/qir/test_pytket_qir_rangepredicate.ll
```diff
@@ -11,32 +11,26 @@ define void @main() #0 {
 entry:
```
cqc-melf, line 11: This file is part of the tests, it is just used to compared the generated qir to what we are expecting

CalMacCQ (approved):

Issue #50, opened 2023-07-19
Remove empty else branch generation


peter-campora:
Conditional pytket gates like:
```python
circuit.X(data[0], condition=reg_eq(syndrome, 1))
circuit.X(data[0], condition=reg_eq(syndrome, 2)) 
```

cause unnecessary empty branches to be generated in the LLVM like:

```llvm
then7:                                            ; preds = %continue6
  call void @__quantum__qis__phasedx__body(double 0x400921FB54442D18, double 0.000000e+00, %Qubit* inttoptr (i64 7 to %Qubit*))
  br label %continue9

else8:                                            ; preds = %continue6
  br label %continue9

continue9:                                        ; preds = %else8, %then7
  %88 = call i1 @read_bit_from_reg(i64 %2, i64 5)
  br i1 %88, label %then10, label %else11
```

It would help program performance if these empty branches were removed.

cqc-melf:
We could use this with a llvm pass after the qir is generated.

cqc-melf:
@peter-campora is this still something we want to to here? From my understanding we could close this issue?

peter-campora:
Done

PR #56, opened 2023-07-21
try to remove empty blocks from qir with new optimisation option

cqc-melf:
Pyqir creates a lot of empty bocks which are not doing anything, There was a request that we should try to remove them. I have not found a way to remove them in the pyqir generation. See https://github.com/CQCL/pytket-qir/issues/50

### DIFF
--- a/tests/conditional_test.py
+++ b/tests/conditional_test.py
@@ -172,7 +172,8 @@ def test_pytket_qir_conditional_7() -> None:
 
 
 def test_pytket_qir_conditional_8() -> None:
-    c = Circuit(4)
+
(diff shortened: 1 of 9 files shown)


tests/conditional_test.py
```diff
@@ -173,7 +173,7 @@ def test_pytket_qir_conditional_7() -> None:
 
 def test_pytket_qir_conditional_8() -> None:
 
-    c = Circuit(4)
+  c = Circuit(4)
```
cqc-melf, line 176:
```suggestion
    c = Circuit(4)

PR #57, opened 2023-07-30
Prefer `init_reg` over `reg2var` for empty classical registers

qartik:
`reg2var` is too verbose for empty registers. I would like to propose `init_reg` instead which  is not only concise but preserves valuable register size information.

I also added a `Makefile` and config files for pre-commit and ruff. While doing the latter, I ended up fixing some typos and improving some code.

I am happy to hear other suggestions such as perhaps using `init_reg` to completely replace `reg2var` by passing an initial value in the form of a decimal i64.


tests/conversion_test.py
```diff
@@ -191,21 +190,6 @@ def test_pytket_qir_10() -> None:
...
-
-    c.add_c_copyreg(a, b)
-
-    result = pytket_to_qir(c, name="test_pytket_qir_10", qir_format=QIRFormat.STRING)
-
-    check_qir_result(
-        result, "test_pytket_qir_10"
-    )  # should be identical to the testcase above
-
-
```
qartik, line 208: Note I deleted this test because with `init_reg`, the output is different from `test_pytket_qir_10`

tests/conditional_test.py
```diff
@@ -119,20 +133,28 @@ def test_pytket_qir_conditional_v() -> None:
...
-        circ, name="test_pytket_qir_conditional_v", qir_format=QIRFormat.STRING
+    result = pytket_to_qir(
+        circ,
+        name="test_pytket_qir_conditional_5",
+        qir_format=QIRFormat.STRING,
     )
 
-    check_qir_result(result, "test_pytket_qir_conditional_v")
-
-    check_qir_result(result_2, "test_pytket_qir_conditional_v")
```
qartik, line 135: Also refactored this test into its own test: `test_pytket_qir_conditional_5`

.github/workflows/docs/build-docs
```diff
@@ -28,7 +28,7 @@ def remove_dir(dirpath):
 
 
 def fix_links(filepath):
-    with open(filepath, "r", encoding="utf8") as f:
```
cqc-melf, line 31: Why is the "r" removed?

.github/workflows/docs/conf.py
```diff
@@ -1,5 +1,3 @@
-# -*- coding: utf-8 -*-
```
cqc-melf, line 1: Why is this removed?

Makefile
```diff
@@ -0,0 +1,18 @@
+.PHONY: build tests install clean dev
```
cqc-melf, line 1: Can you please add something to the readme how this could be used?

README.md
```diff
@@ -1,18 +1,16 @@
-# Pytket Extensions
-
-# this is currently work in progress
```
cqc-melf, line 3: I would like to keep this in the first two lines, to make sure that people see that this is not yet recommended to been used?

README.md
```diff
@@ -32,7 +30,17 @@ There is also a Slack channel for discussion and support. Click [here](https://t
 
 ## Development
 
-To install this extension in editable mode, simply change to this directory, and run:
+First setup your virtual environment and install required dependencies:
+
+```sh
+python -m venv .venv
```
cqc-melf, line 36: It a virtual environment required in this way? Would it not be possible to run this in the local python?

pytket/qir/conversion/conversion.py
```diff
@@ -179,8 +176,17 @@ def __init__(
...
             pyqir.FunctionType(
-                pyqir.IntType(self.module.module.context, qir_int_type),
-                [pyqir.IntType(self.module.module.context, 1)] * qir_int_type,
+                pyqir.IntType(self.module.module.context, qir_int_size),
+                [pyqir.IntType(self.module.module.context, 1)] * qir_int_size,
+            ),
+        )
+
+        # i64 init_reg(i64 size), create an empty classical register of size bits
+        self.init_reg = self.module.module.add_external_function(
```
cqc-melf, line 185: My understanding is that the default value is always 0 of this register? is that right?

pytket/qir/conversion/conversion.py
```diff
@@ -394,33 +413,36 @@ def _to_qis_bits(self, args: List[Bit]) -> Sequence[Value]:
...
     def _reg2ssa_var(self, bit_reg: BitRegister, int_size: int) -> Value:
         """Convert a BitRegister to an SSA variable using pyqir types."""
-        reg_name = bit_reg[0].reg_name
-        if (
-            reg_name not in self.ssa_vars.keys()
-        ):  # Check if the register has been previously set.
+        reg_name = bit_reg.name
+        if reg_name not in self.ssa_vars.keys():
             # Check if the register has been previously set. If not, initialise to 0.
+            empty_reg = False
```
cqc-melf, line 419: Why is it important to handle this as a special case?
What is stopping us from just not converting an empty register?

pytket/qir/conversion/conversion.py
```diff
@@ -629,10 +658,9 @@ def condition_block() -> None:
                     )
 
             elif isinstance(op, WASMOp):
-                raise ValueError("WASM not supported yet")
+                raise TypeError("WASM not supported yet")
```
cqc-melf: I think this is an error in the input and so it should be a value error?

ruff.toml
```diff
@@ -0,0 +1,41 @@
+target-version = "py39"
```
cqc-melf, line 1: Can you add something to the readme how to use this file?

tests/conditional_test.py
cqc-melf: So they are not checked to be identical anymore? What it the reason for that?

cqc-melf (commented):
Hi @qartik
Than you for the suggestions :) 

Have you discussed the changes with @peter-campora and if this will be supported by the backend in the same way?

I am not really sure if this is the right thing to do, I would assume that it might be better to keep the register size outside of qir. It is not really clear to me what the expected returns type are if we execute operations on registers with different sizes. Can you maybe give some more details on that?

We usually try to separate clean up / ci / infrastructure changes from new features. Do you think it would be possible to split this up into two PRs?

ruff.toml
qartik: This file is a configuration file for [ruff](https://github.com/astral-sh/ruff) which is an extremely fast linter for Python. It is automatically called by pre-commit (right before one commits, assuming we have issued `pre-commit install`), which I included in the README.

Documentation for the rules in this file is at https://beta.ruff.rs/docs/rules/

tests/conditional_test.py
qartik: Because the first circuit is `Circuit(2, 3)` and the second `Circuit(2, 2)`, ie, they have different sizes for the classical register.

qartik:
> Have you discussed the changes with @peter-campora and if this will be supported by the backend in the same way?

Yes, once this new `init_reg` API is stabilized and accepted here, I will ensure the backend supports it as well.

> I am not really sure if this is the right thing to do, I would assume that it might be better to keep the register size outside of qir. It is not really clear to me what the expected returns type are if we execute operations on registers with different sizes. Can you maybe give some more details on that?

The return types continue to be `i64`. Since the backend only consumes QIR, I believe that's the only place to store register size information, though I am open to alternatives if they exist. Think of the first argument to `init_var` as an additional metadata for the backend to distinguish it from real integers from other frontends such as Q# -> QIR. This extra information may also be used to decide do other optimizations at the backend (if needed).

> We usually try to separate clean up / ci / infrastructure changes from new features. Do you think it would be possible to split this up into two PRs?

I can do that. But my logical changes are on top of the linting changes such as black pass, which seemed like was not run in quite some time. So if the linting changes look okay, I can create a PR for those first.

.github/workflows/docs/build-docs
qartik: It's the default option and [hence unneeded](https://beta.ruff.rs/docs/rules/redundant-open-modes/).

.github/workflows/docs/conf.py
qartik: utf-8 is the default and [hence unneeded](https://beta.ruff.rs/docs/rules/utf8-encoding-declaration/).

README.md
qartik: Yes, but it is certainly better to use a virtual environment to keep packages and dev environment isolated.

pytket/qir/conversion/conversion.py
qartik: That's correct. Though I am open to suggestions for the API such as passing in the initial value to the function. Let's discuss in our meeting today.

pytket/qir/conversion/conversion.py
qartik: I am reverting this change, though I had changed it because it was a failure of `isinstance` check which is about types not values.

pytket/qir/conversion/conversion.py
qartik: I did not understand this question fully. In the existing code there is a check for whether the classical register has been written into (set), so I am only introducing `init_reg` for when that's not the case, ie, for an empty register.

We could completely remove `reg2var` and use `init_reg` (or something with an even better name) to pass in a value other than zero. This is something we can discuss.

.github/workflows/docs/conf.py
cqc-melf: Is that the case for all os?

README.md
cqc-melf: Okay, in that case I would only say that in a sentence and remove that from the code, it just confuses people that uses something else for environments

ruff.toml
cqc-melf: If we want to change the lint standard from the used pylint to ruff we should change the lint run on the ci as well. We should make sure that this is done on every PR.

cqc-melf:
> > We usually try to separate clean up / ci / infrastructure changes from new features. Do you think it would be possible to split this up into two PRs?
> 
> I can do that. But my logical changes are on top of the linting changes such as black pass, which seemed like was not run in quite some time. So if the linting changes look okay, I can create a PR for those first.

The current used pylint and black is run on every PR, for me it looks like that ruff is just following a different convention.
I think it would be good to move that to a different PR. If there are benefits for changing the package we are using for that, we could think about changing that for our other pytket extensions as well.
This should be updated in the readme as well that we are now using ruff, see https://github.com/CQCL/pytket-qir#formatting

I think it would be good to have an additional PR for changing the ci to use ruff and run ruff once on the code to have a clean commit on develop which only contains the changes related to that. (And doing the precommit, makefile, ... changes in an additional separate PR after that.)

I am happy to help you with that, splitting this up, might be a little bit more work.

.github/workflows/docs/conf.py
qartik: That's my understanding, see https://stackoverflow.com/a/14083123/1167061

README.md
qartik: Perhaps, we should settle on a tool then? Do you prefer something else?

I am primarily following the convention from CQCL's `pytemplate` project that I have also contributed to. See https://github.com/CQCL/pytemplate#installation

ruff.toml
qartik: Good idea, I was not able to run the CI on my own earlier, but I can give it a try.

To be clear, ruff currently doesn't fully replace Pylint in general, but for this project's needs it might. You seem to be using a pretty permissive pylint configuration.

qartik:
> The current used pylint and black is run on every PR, for me it looks like that ruff is just following a different convention.
> I think it would be good to move that to a different PR. If there are benefits for changing the package we are using for that, we could think about changing that for our other pytket extensions as well.
> This should be updated in the readme as well that we are now using ruff, see https://github.com/CQCL/pytket-qir#formatting

It's possible that since I used the most recent version of black, it did things differently. But it seems compatible with what was on the CI.

BTW, ruff doesn't format things. Most of the formatting was performed by black. But I will double check that when I create a separate PR.

cqc-melf:
If the changes from you are done by just updating the black version we could do that as well and just add ruff in addition to that?

README.md
cqc-melf: For the pytket extension we have used the pytket-quantinuum repo as a template, we don't have precommit hocks in there at the moment, but everything else should be similar. It has some additional things to be compatible with the other public documentation

qartik:
> BTW, ruff doesn't format things. Most of the formatting was performed by black. But I will double check that when I create a separate PR.

Indeed, I was mistaken, a lot of changes came from my initial trials with `ruff`. But I was more careful in the separate PR for linting/automation improvements: https://github.com/CQCL/pytket-qir/pull/58

PR #58, opened 2023-08-01
Add pre-commit with black, ruff and pylint automation

qartik:
Also:
- improve README
- update package versions
- add Makefile
- resolve typos and linting issues reported by above tools
- remove lark warnings from pytest output


.github/workflows/docs/build-docs
```diff
@@ -28,7 +28,7 @@ def remove_dir(dirpath):
 
 
 def fix_links(filepath):
-    with open(filepath, "r", encoding="utf8") as f:
```
cqc-melf, line 31: I am still not sure why this change is made, have you done this manually or is ruff doing this changes? I have done a PR with only updating black and that is not doing this change.

.github/workflows/docs/conf.py
```diff
@@ -1,5 +1,3 @@
-# -*- coding: utf-8 -*-
```
cqc-melf, line 1: See above

.github/workflows/docs/conf.py
```diff
@@ -92,7 +91,8 @@
 def add_reference(
     app: Sphinx, env: BuildEnvironment, node: Element, contnode: TextElement
 ) -> Optional[nodes.reference]:
-    # Fix references in docstrings that are inherited from the base pytket.backends.Backend class.
+    # Fix references in docstrings that are inherited
```
cqc-melf, line 94: See above

.github/workflows/docs/build-docs
qartik: Same answer as before: https://github.com/CQCL/pytket-qir/pull/57#discussion_r1280597356 Ruff does it with its implementation of the pyupgrade linter.

Copying over:

> It's the default option and [hence unneeded](https://beta.ruff.rs/docs/rules/redundant-open-modes/).

.github/workflows/docs/conf.py
qartik: See https://github.com/CQCL/pytket-qir/pull/57#discussion_r1280599374 and later comments in the thread.

.github/workflows/docs/conf.py
qartik: This change is manual to conform to the line width limit.

.github/workflows/docs/build-docs
cqc-melf: I am fine with doing this change, maybe my question was not that clear. Had ruff done this automatically or have you done this change manually? And if ruff has done that, with which command run?

README.md
```diff
@@ -32,12 +30,32 @@ There is also a Slack channel for discussion and support. Click [here](https://t
 
 ## Development
 
-To install this extension in editable mode, simply change to this directory, and run:
+First setup your virtual environment and install required dependencies:
+
+```sh
+python -m venv .venv
```
cqc-melf, line 36: See https://github.com/CQCL/pytket-qir/pull/57#discussion_r1280635199

I think it would be good to do something like

```
First install required dependencies:

```sh
pip install -U pip setuptools
pip install build pre-commit pytest wheel mypy~=1.4 black~=23.7 pylint~=2.17 ruff==0.0.282
pre-commit install
```

Then install this extension in editable mode, simply change to this directory, and run:

.github/workflows/docs/build-docs
qartik: `ruff check .` shows all the lint violations. Some of those can be automatically corrected by ruff using `ruff check . --fix`. But in this PR, we also have `pre-commit` set up in such a way that `ruff check --fix` will be automatically called on all files that are being added/modified in the commit. The last sentence is True only if `pre-commit install` has been issued once in your local clone once as mentioned in the README in the development steps. I am happy to show a demo if desired.

README.md
qartik: What's the argument against providing steps for creating a virtual environment for beginners? I like to provide a self-contained set of steps (especially those that I can copy and paste myself without having to think). People who are already familiar with virtual environments and prefer setting them up any other way can just skip copying the first two lines, no?

cqc-melf (approved):

PR #59, opened 2023-08-01
Prefer `init_reg` over `reg2var` for empty classical registers

qartik:
As discussed in #57 and later.

I would like to use this PR for stabilizing the design of the API. For example, here are a couple options:

1. remove `reg2var` completely and keep `init_reg` as is
1. use `reg2var` for only creating registers with initial values and `init_reg` for empty cregs
1. add an additional parameter to `init_reg` for an initial value and remove `reg2var`.

PS: Please ignore changes from #58, see https://github.com/CQCL/pytket-qir/pull/59/commits/b3a04dd30c79ce024cd3c22b2f2682395c3bb52d for the changes introduced in this PR.


qartik:
Also see another option #61

qartik:
Closing in favor of #61

PR #63, opened 2023-08-08
fix python version

cqc-melf:
Looks like specifying a python version here make some problems when this version is not used locally. I would suggest that we should remove this.


qartik:
I am fine with this removal. Just out of curiosity, what error/problem do you see?

cqc-melf:
The black run is crashing if the local python version does not match the version given in the file

qartik (approved):

Issue #65, opened 2023-08-08
Consider removing `pyqir_0_6_compatibility` code


qartik:
We explicitly require pyqir 0.8.2 at
https://github.com/CQCL/pytket-qir/blob/6b500b808831a545ada0670c5a2ae61ea7bcb227/setup.py#L49

So the bit below and other `pyqir_0_6_compatibility` code is unneeded.
https://github.com/CQCL/pytket-qir/blob/6b500b808831a545ada0670c5a2ae61ea7bcb227/pytket/qir/conversion/api.py#L88-L129

qartik:
@cqc-melf any objections? I can go ahead and remove this dead code if you agree.

cqc-melf:
We have added this in the past because we wanted to be compatible with older versions used on your side, if you don't need this it can be removed. Are you using now 0.8.x as well?

qartik:
Yes, we are. There is no support for PyQIR below 0.8.2 in the backend compiler, AFAIK.

Just to confirm, @peter-campora do you think old versions are still in use and may require compatibility with pytket-qir?

Update (2023-08-15): We agreed that it can be removed.

cqc-melf:
solved by #66

PR #66, opened 2023-08-16
add wasm support to the conversion

cqc-melf:


cqc-melf:
Solved https://github.com/CQCL/pytket-qir/issues/65 and adds the wasm support

pytket/qir/conversion/api.py
```diff
@@ -75,49 +82,22 @@ def pytket_to_qir(
...
-            raise ValueError(
-                """The qir optimised for pyqir 0.6 can only contain
-one classical register"""
-            )
+    if wfh is not None:
+        wasm_dict: dict[str, str] = qir_generator.get_wasm_sar()
 
         initial_result = str(populated_module.module.ir())  # type: ignore
 
-        initial_result = (
```
cqc-melf, line 97: It is probably easier to understand this when looking at the individual commits

pytket/qir/conversion/conversion.py
```diff
@@ -244,6 +251,42 @@ def __init__(
...
+                parametertype = [self.qir_int_type] * wfh._functions[fn][0]
+                if wfh._functions[fn][1] == 0:
+                    returntype = pyqir.Type.void(self.module.module.context)
+                    wasm_func_interface += "void "
+                elif wfh._functions[fn][1] == 1:
+                    returntype = self.qir_int_type
+                    wasm_func_interface += f"i{self.int_type_str} "
+                else:
+                    raise ValueError(
+                        "wasm function which return more than"
```
cqc-alec, line 268: I guess we should make an issue for this?

cqc-alec (approved):

pytket/qir/conversion/conversion.py
cqc-melf: Yes I will do that

cqc-melf: Done at https://github.com/CQCL/pytket-quantinuum/issues/201
I have added that to pytket-quantinuum because this is an issues that needs to be solved for qasm, qir and on the hardware side

PR #68, opened 2023-08-16
fix wasm issue, clean up

cqc-melf:


cqc-melf:
I have found one more issue in the generated qir, there was in `i` to much in the search and replace.

cqc-alec (approved):

PR #71, opened 2023-08-17
fix documentation

cqc-melf:


cqc-melf:
Small documentation fix

qartik (approved):

pytket/qir/conversion/api.py
```diff
@@ -55,9 +55,9 @@ def pytket_to_qir(
     :type name: str
     :param qir_format: format of the generated qir, default value is binary
     :type qir_format: QIRFormat
-    :param pyqir_0_6_compatibility: converts the output to be compatible with
-        pyqir 0.6, default value false
-    :type pyqir_0_6_compatibility: bool
+    :param wfh: wasm file handler used when creating the circuit.
+      Only needed when there are wasm calls in the circuit.
+    :type wfh: pytket.wasm.WasmFileHandler
```
```suggestion
    :type wfh: Optional[pytket.wasm.WasmFileHandler]
```

cqc-alec (dismissed):

qartik:
Ah, good catch. I wonder if the documentation generator can just pick the types from the annotations instead of us having to write them and potentially making copy/paste errors.

cqc-alec:
> Ah, good catch. I wonder if the documentation generator can just pick the types from the annotations instead of us having to write them and potentially making copy/paste errors.

I think it's not necessary to include these `:type:` annotations anyway, as the types are annotated and visible in the documentation.

cqc-melf:
I have now removed all the types

qartik (approved):
Thanks for the quick update!

PR #82, opened 2023-10-17
mypy fix

cqc-melf:


pytket/qir/conversion/conversion.py
```diff
@@ -508,8 +508,8 @@ def _get_ssa_from_cl_reg_op(
             return output_instruction  # type: ignore
         elif type(reg) == BitRegister:
             return self._get_i64_ssa_reg(reg.name)
-        elif type(reg) == int:
-            return pyqir.const(self.qir_int_type, reg)
+        elif type(reg) == int:  # type: ignore
```
cqc-alec: Is this condition possible? (If so it looks like the type annotation of `reg` above must be wrong?

cqc-melf: Oh yes, fixed now. See [f13a65f](https://github.com/CQCL/pytket-qir/pull/82/commits/f13a65f64cb56c75ed8824a58a406b4313606e4c)

cqc-alec (approved):

PR #61, opened 2023-08-07
update registerhandling to use i1* pointer

cqc-melf:

PR #87, opened 2023-10-24
Feature/measure to reg

cqc-melf:
Not sure what the name is er agreed on, happy to update this.

### DIFF
--- a/pytket/qir/conversion/conversion.py
+++ b/pytket/qir/conversion/conversion.py
@@ -202,7 +202,7 @@ class QirGenerator:
             ),
         )
 
-        # void mz_to_creg(qubit, i1* creg, int)
+        # void mz_to_creg(qubit, i1* creg, int creg_index)
         self.mz_to_creg = self.module.module.add_external_function(
             "mz_to_creg",
             pyqir.FunctionType(
--- a/pytket/qir/conversion/conversion.py
+++ b/pytket/qir/conversion/conversion.py
@@ -202,9 +202,10 @@ class QirGenerator:
             ),
         )
 
-        # void mz_to_creg(qubit, i1* creg, int creg_index)
-        self.mz_to_creg = self.module.module.add_external_function(
-            "mz_to_creg",
+        # void mz_to_creg_bit(qubit, i1* creg, int creg_index)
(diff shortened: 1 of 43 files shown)


pytket/qir/conversion/conversion.py
```diff
@@ -202,6 +202,19 @@ def __init__(
             ),
         )
 
+        # void mz_to_reg(qubit, i1* creg, int)
```
qartik: Could you write the signature here in full? What's the `int` type referring to?

qartik (changes requested):

qartik:
> Not sure what the name is er agreed on

Given the interface in https://github.com/CQCL/pytket-qir/pull/61#pullrequestreview-1567770240 perhaps a better name might be `measure_to_creg` or `mz_to_creg` in short?

pytket/qir/conversion/conversion.py
cqc-melf: done

cqc-melf:
> > Not sure what the name is er agreed on
> 
> Given the interface in [#61 (review)](https://github.com/CQCL/pytket-qir/pull/61#pullrequestreview-1567770240) perhaps a better name might be `measure_to_creg` or `mz_to_creg` in short?

I have done a rename to `mz_to_creg`

qartik:
Hi @cqc-melf, it's not clear to me from a quick look whether `mz_to_creg` is supposed to measure a single qubit or a register of qubits or both. Can you clarify?

If it's a single bit (that I suspect from the third argument of creg index), then the name could be misleading. In that case, I'd prefer it be called `mz_to_creg_bit`.

cqc-melf:
Yes the plan is to measure one qubit to the bit entry of a register, I have added an additional comment and renamed like suggested

qartik (commented):
LGTM, but I will let @peter-campora chip in as well before approval.

cqc-alec (approved):

peter-campora (approved):
`mz_to_creg_bit` is a name that works for me. Will add support for it in L3 soon!

qartik (approved):

PR #120, opened 2024-01-29
Feature/nix support

jake-arkinstall:
# Description

Added nix support, The nix flake fetches the pytket flake to fetch pytket, and uses a custom derivation to generate pyqir. mypy and pytest are run as part of `nix flake check`, which is performed in CI (currently set to every sunday at 2am).

At the moment, tket and pytket will be built. The speed of performing the flake check will be improved significantly if we use cachix.

To try out this flake:
```
nix develop github:CQCL/pytket-qir/feature/nix-support
```

Then invoke python. `pytket.qir` is importable and working, e.g.:
```
$ nix develop github:CQCL/pytket-qir/feature/nix-support

$ python3
Python 3.11.7 (main, Dec  4 2023, 18:10:11) [GCC 13.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from pytket.circuit import Circuit
>>> from pytket.qir.conversion.api import QIRFormat, pytket_to_qir
>>> c = Circuit(2)
>>> c.H(0)
[H q[0]; ]
>>> c.H(1)
[H q[0]; H q[1]; ]
>>> c.CX(0, 1)
[H q[0]; H q[1]; CX q[0], q[1]; ]
>>> c.H(0)
[H q[0]; H q[1]; CX q[0], q[1]; H q[0]; ]
>>> c.H(1)
[H q[0]; H q[1]; CX q[0], q[1]; H q[0]; H q[1]; ]
>>> print(pytket_to_qir(c, name="example", qir_format=QIRFormat.STRING))
; ModuleID = 'example'
source_filename = "example"

%Qubit = type opaque
%Result = type opaque
; ... and so on
```

After the merge of this PR, the development shell will be accessible with `nix develop github:CQCL/pytket-qir`
# Checklist

- [ ] I have made corresponding changes to the public API documentation. (N/A)
- [ ] I have updated the changelog with any user-facing changes. (N/A)

### DIFF
--- a/.github/workflows/build_with_nix.yml
+++ b/.github/workflows/build_with_nix.yml
@@ -1,8 +1,8 @@
 name: build with nix
 on:
-  schedule:
-    # 02:00 every Sunday morning
-    - cron: '0 2 * * 0'
+  pull_request:
+    branches:
+      - main
   workflow_dispatch: {}
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
@@ -11,10 +11,14 @@ jobs:
   build_and_test:
     strategy:
       matrix:
-        os: ['ubuntu-22.04', 'macos-13-xlarge']
+        os: ['ubuntu-22.04', 'macos-14']
     runs-on: ${{matrix.os}}
     steps:
     - uses: actions/checkout@v4
     - uses: cachix/install-nix-action@v25
-    - name: Build and test tket
+    - uses: cachix/cachix-action@15
+      with:
+        name: tket
+        signingKey: ${{ secrets.CACHIX_AUTH_TOKEN }}
+    - name: Build and test pytket-qir
       run: nix flake check -L
--- a/.github/workflows/build_with_nix.yml
+++ b/.github/workflows/build_with_nix.yml
@@ -16,7 +16,7 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - uses: cachix/install-nix-action@v25
-    - uses: cachix/cachix-action@15
+    - uses: cachix/cachix-action@v15
       with:
         name: tket
         signingKey: ${{ secrets.CACHIX_AUTH_TOKEN }}
--- a/.github/workflows/build_with_nix.yml
+++ b/.github/workflows/build_with_nix.yml
@@ -1,0 +1,20 @@
+name: build with nix
+on:
+  schedule:
+    # 02:00 every Sunday morning
+    - cron: '0 2 * * 0'
+  workflow_dispatch: {}
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+jobs:
+  build_and_test:
+    strategy:
+      matrix:
+        os: ['ubuntu-22.04', 'macos-13-xlarge']
+    runs-on: ${{matrix.os}}
+    steps:
+    - uses: actions/checkout@v4
+    - uses: cachix/install-nix-action@v25
+    - name: Build and test tket
+      run: nix flake check -L
--- a/flake.lock
+++ b/flake.lock
@@ -5,11 +5,11 @@
         "systems": "systems"
       },
       "locked": {
-        "lastModified": 1705309234,
-        "narHash": "sha256-uNRRNRKmJyCRC/8y1RqBkqWBLM034y4qN7EprSdmgyA=",
+        "lastModified": 1710146030,
+        "narHash": "sha256-SZ5L6eA7HJ/nmkzGG7/ISclqe6oZdOZTNoesiInkXPQ=",
         "owner": "numtide",
         "repo": "flake-utils",
-        "rev": "1ef2e671c3b0c19053962c07dbda38332dcebf26",
+        "rev": "b1d9ab70662946ef0850d488da1c9019f3a9752a",
         "type": "github"
       },
       "original": {
@@ -23,11 +23,11 @@
         "systems": "systems_2"
       },
       "locked": {
-        "lastModified": 1694529238,
-        "narHash": "sha256-zsNZZGTGnMOf9YpHKJqMSsa0dXbfmxeoJ7xHlrt+xmY=",
+        "lastModified": 1710146030,
+        "narHash": "sha256-SZ5L6eA7HJ/nmkzGG7/ISclqe6oZdOZTNoesiInkXPQ=",
         "owner": "numtide",
         "repo": "flake-utils",
-        "rev": "ff7b65b44d01cf9ba6a71320833626af21126384",
+        "rev": "b1d9ab70662946ef0850d488da1c9019f3a9752a",
         "type": "github"
       },
       "original": {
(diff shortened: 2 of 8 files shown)


.github/workflows/build_with_nix.yml
```diff
@@ -0,0 +1,20 @@
+name: build with nix
+on:
+  schedule:
```
cqc-melf: Can you give some details why we would not want to run this on PRs? In case of changes we would like to know that this fails?

flake.lock
cqc-melf, line 1: Is this a file we want to have in the git?

nix-support/pyqir.nix
```diff
@@ -0,0 +1,45 @@
+self: super:
+let
+  lib = super.lib;
+  llvm = super.llvm_14;
+  llvm-v-major = lib.versions.major llvm.version;
+  llvm-v-minor = builtins.substring 0 1 (lib.versions.minor llvm.version);
+  pyqir-version = "0.10.0";
```
cqc-melf, line 7: Is it possible to import the version from the python packages here, or do we now need to update this in two places?

cqc-melf (commented):
Thank you for the suggestion!
I am not familiar with nix, so I have added a few questions / comments

.github/workflows/build_with_nix.yml
jake-arkinstall: Ah, the moment we don't have a Cachix subscription, so we sint have a central store to place things that we have built before. This includes tket - so when this CI runs, it will need to build tket first.

Once we have a caching subscription, the latest tket build will already be available, so building and testing pytket-qir will just be a case of bundling the python package and running checks on it. Then, if we store the result in cachix, anyone can load up a pre-built environment with tket, pytket and pytket-qir very quickly.

flake.lock
jake-arkinstall: Yes. The flake.lock pins the nix flake dependencies to well-defined versions, which helps keep the repository and users of the repository in sync.

It's feasible that a future commit to nixpkgs e.g. breaks mypy, or updates scipy with a breaking change. By keeping the flake.lock file in the repository, maintainers of this repository get to decide when nixpkgs, tket etc can get updated, by pushing changes created by invoking `nix flake update`.

nix-support/pyqir.nix
jake-arkinstall: In this case it is customary to provide the exact git revision to pull, then specify the build process, as pyqir isn't available on nixpkgs.

I can update it to scan setup.py for a specific version, though if it changes to using a loose version match then the process will fail.

Though I will add that I am committed to supporting the nix support for this repository and others, so if a change breaks the nix support, or if a version update is required, I'm absolutely happy to action it.

.github/workflows/build_with_nix.yml
cqc-melf: Thank you for all the details!

nix-support/pyqir.nix
cqc-melf: I case I am doing the update, is there anything else I need to do besides updating the version in this file as well?

cqc-melf (commented):
Do you have plans to add nix for the other extensions as well?

cqc-melf:
And it would be nice if you could add something to the changelog?

jake-arkinstall:
For some reason my github app isnt letting my reply in thread, so:

> I case I am doing the update, is there anything else I need to do besides updating the version in this file as well?

If the new version requires to change to the setup process, then this is my usual update process:

1. Reset the hash to an empty string: ""
2. Update the version number to an available tag of the repository
3. Run "nix build"

The nix build log will then emit a warning along the lines of:
"Empty hash provided, assuming sha256 AAAAAAAA...."

Once it has built the package it will calculate the actual hash. It will emit an error along the lines of:
```
Incorrect hash.
    Specified: AAAAAAAA.....
    Got: "ABCDEF1234....."
```

What I do then is add the provided hash back into the nix file and run nix build again.

It should immediately notice that it has already seen a derivation with this source, build instructions, and output hash, and resolve to the package it previously built.

The hash thing is annoying but extremely useful. If Nix can verify the hash then it knows if it has seen it before and lets it skip steps.

Likewise, subsequent runs on that machine will serve that result immediately. If we get a cachix subscription, we can upload the resulting derivation there, and any user that needs it (as specified by the matching source, build instructions and output hash) will be offered a ready-to-go downloadable package instead of going through the build and test process.

> Will you do the same with other packages?

Indeed! I was asked to look at pytket-quantinuum and pytket-qir is a dependency of it (in the testing pipeline) so I started here.

This, with the above about the caching possibilities, is quite exciting.

Cachix is when the real benefit of the nix integration will kick in: users will be able to run `nix develop github:CQCL/pytket-foo` on any Linux or Mac Silicon machine and everything from symengine to tket, pyket to pyqir, and everything they need, will be served over a download with no additional building or configuration required. One command and after a few seconds they'll be ready to go.

It doesn't stop at environments, but can also launch tools. So we could possibly have a one-liner to launch a Jupiter notebook pre-filled with demo code, with the environment all loaded up, ready to showcase. But this is all future ifs and maybes, and I don't want to push it in a direction you don't want to take it in.

cqc-melf:
Thank you for all the information! I think it would be helpful if you could just copy this into a readme file and add this to the repo as well. Just to make sure this is documented somewhere. 

Happy to review your PR on pytket-quantinuum as well.

cqc-melf:
Maybe it would be good if we could add this to pytket-qiskit as well? As they are both the most used extensions?

jake-arkinstall:
That is my next challenge after I've got pytket-quantinuum in.

There are qiskit libraries already available on the nix store, however they
are marked as broken due to conflicts with symengine. So the qiskit
integration PR might be a bit more substantial of a PR than this one,
depending on how it goes.

cqc-melf:
@jake-arkinstall is this ready for re-review?

jake-arkinstall:
> @jake-arkinstall is this ready for re-review?

Yes, it is! Although as some movement looks to have been made on the cachix side it might be worth waiting for that to land in tket itself, as it'll speed things up considerably (this repo can then rely on the tket's cached outputs instead of forcing a build).

cqc-alec:
Please rebase and retarget to `main` which is now the default branch.

jake-arkinstall:
I have rebased to `main`, and I have also updated the flake and workflow to utilise the TKET cachix (binary cache).

For the workflow to succeed, CACHIX_AUTH_TOKEN will need to be available in the workflow environment, as in the tket repository.

I have also run `nix flake check` locally (x86_64-linux) and everything passes.

jake-arkinstall:
(Additionally, I set nixpkgs to follow TKET's - this will help with possible cache misses if the two disagree on nixpkgs. As more weight is on TKET's build, doing it this way around seems to make the most sense)

cqc-alec:
I added `CACHIX_AUTH_TOKEN` to the repo.

cqc-alec:
Closing and reopening to trigger CI...

.github/workflows/build_with_nix.yml
```diff
@@ -0,0 +1,24 @@
...
+jobs:
+  build_and_test:
+    strategy:
+      matrix:
+        os: ['ubuntu-22.04', 'macos-14']
+    runs-on: ${{matrix.os}}
+    steps:
+    - uses: actions/checkout@v4
+    - uses: cachix/install-nix-action@v25
+    - uses: cachix/cachix-action@15
```
cqc-alec:
```suggestion
    - uses: cachix/cachix-action@v15
```

cqc-melf:
@jake-arkinstall are you still working on this, or can we close it?

jake-arkinstall:
*pew pew pew*

PR #131, opened 2024-04-16
remove copyright from documentation

cqc-melf:
# Description

Please summarise the changes.

# Related issues

# Checklist

### DIFF
--- a/.github/workflows/docs/build-docs
+++ b/.github/workflows/docs/build-docs
@@ -75,8 +75,6 @@ def build_module_docs():
             "-D",
             f"project=pytket-{MODULE}",
             "-D",
-            f"copyright={datetime.date.today().year} Cambridge Quantum Computing",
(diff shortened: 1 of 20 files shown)


.github/workflows/docs/build-docs
```diff
@@ -75,8 +75,6 @@ def build_module_docs():
             "-D",
             f"project=pytket-{MODULE}",
             "-D",
-            f"copyright={datetime.date.today().year} Cambridge Quantum Computing",
```
qartik, line 78: Why remove, and not just update to Quantinuum?

cqc-melf: We are not publishing the docs from this repo anymore, what we would set here would stop us from setting it later at a central point for all extensions.

cqc-alec (approved):

PR #147, opened 2024-06-27
speedup and cleanup

cqc-melf:
# Description

Speed up the conversion
update pytket version
add testcase for invalid register

# Checklist

- [x] I have made corresponding changes to the public API documentation.
- [x] I have updated the changelog with any user-facing changes.


pytket/qir/conversion/conversion.py
cqc-melf, line 1: The github diff of this file is a bit cryptic. The main changes are that I have split up the circuit_to_module function which was called recursively into three functions for circuit, subcircuit and command conversion.

cqc-melf: Happy to give more details if that would be helpful

tests/qasm_test.py
```diff
@@ -57,6 +57,45 @@ def test_pytket_qir_qasm_ii() -> None:
         check_qir_result(result, "test_pytket_qir_qasm")
 
 
+def test_pytket_qir_qasm_long() -> None:
+    try:
```
cqc-alec: Not sure if you intended to include this test, which will never run on CI? Better to check in the circuit files with a separate test script that can be run locally; otherwise this is useless unless one happens to have the files.

pytket/qir/conversion/api.py
```diff
@@ -46,6 +52,8 @@ def pytket_to_qir(
     wfh: Optional[wasm.WasmFileHandler] = None,
     int_type: int = 64,
     cut_pytket_register: bool = False,
+    compile_circuit: bool = False,
+    check_input: bool = True,
```
cqc-alec: I don't see the motivation for these arguments. The compliation pass is specific to H-series devices whereas this should eventually be a general QIR converter. What's wrong with the user compiling the circuit in whatever way they want before converting to QIR. Also I think the checks in `check_input` should happen in all cases; why would you want to turn them off?

.gitignore
```diff
@@ -11,3 +11,6 @@ obj
 docs/extensions
 .ipynb_checkpoints
 pytket/qir/_metadata.py
+test_pytket_qir_qasm_long.qasm
```
cqc-alec: See my other comment; if these files are useful for (local) testing they should be checked in.

pytket/qir/conversion/conversion.py
cqc-alec: OK, so the changes in this file are just a refactor, not functional?

cqc-melf: I have also done simplification to get the speed-up, but nothing in the result should have changed.

tests/qasm_test.py
cqc-melf: I am running this test locally, if you think this should be in a separate file and not in the repo I am happy to move it.
(There is a similar case above)

cqc-melf: (The `qec_ii.qasm` file is also not in the repo)

pytket/qir/conversion/api.py
cqc-melf: `check_input` is taking quiet long, if you are sure that your cirucit is working you might want to skip it.

cqc-melf: We had users complain about missing functionality to directly compiler circuit to a gateset which can be converted. The default of `compile_circuit` is false, so this is only done on request. If you think we don't need this, I am happy to remove it again.

.gitignore
cqc-melf: We can't add this file for legal reasons.

cqc-alec: OK, then I think we should remove all references to it. If the only thing it's testing is performance then we can just move it somewhere else; if it's testing any functionality as well then we should construct another (smaller) file to test that.

pytket/qir/conversion/api.py
cqc-alec: > `check_input` is taking quiet long, if you are sure that your cirucit is working you might want to skip it.

If `check_input` is false but some of the checks would fail, what happens? Does the user just get a conversion error further down the line, or does conversion succeed but produce invalid QIR? If the former, I'm fine with that; if the latter, I think that is an issue...

cqc-alec: > We had users complain about missing functionality to directly compiler circuit to a gateset which can be converted. The default of `compile_circuit` is false, so this is only done on request. If you think we don't need this, I am happy to remove it again.

I would rather keep compilation separate, as it's a cleaner division of functions. Also having this option puts a greater testing burden on pytket-qir which is complex enough as it is.

cqc-melf: > > `check_input` is taking quiet long, if you are sure that your cirucit is working you might want to skip it.
> 
> If `check_input` is false but some of the checks would fail, what happens? Does the user just get a conversion error further down the line, or does conversion succeed but produce invalid QIR? If the former, I'm fine with that; if the latter, I think that is an issue...

If the conversion works you will always get valid QIR. If the check would fail, it will fail later on in the conversion, it can sometimes happen that the error message is not as helpful as it would be when doing the check, but I think this is acceptable.

cqc-melf: > > We had users complain about missing functionality to directly compiler circuit to a gateset which can be converted. The default of `compile_circuit` is false, so this is only done on request. If you think we don't need this, I am happy to remove it again.
> 
> I would rather keep compilation separate, as it's a cleaner division of functions. Also having this option puts a greater testing burden on pytket-qir which is complex enough as it is.

Okay, I will remove this option.

cqc-alec: > > > `check_input` is taking quiet long, if you are sure that your cirucit is working you might want to skip it.
> > 
> > 
> > If `check_input` is false but some of the checks would fail, what happens? Does the user just get a conversion error further down the line, or does conversion succeed but produce invalid QIR? If the former, I'm fine with that; if the latter, I think that is an issue...
> 
> If the conversion works you will always get valid QIR. If the check would fail, it will fail later on in the conversion, it can sometimes happen that the error message is not as helpful as it would be when doing the check, but I think this is acceptable.

In that case, I'd recommend removing the option and relying on downstream errors.

.gitignore
cqc-melf: I have removed the tests, they were only for benchmarking.
See @cqc-melf
[remove qasm tests without file](https://github.com/CQCL/pytket-qir/pull/147/commits/58e65966c7cd54d03f3cc787d784228a0c6b3861)
[58e6596](https://github.com/CQCL/pytket-qir/pull/147/commits/58e65966c7cd54d03f3cc787d784228a0c6b3861)

tests/qasm_test.py
cqc-melf: See below

pytket/qir/conversion/api.py
cqc-melf: The compile option is now removed, see @cqc-melf
[remove compiler option from api](https://github.com/CQCL/pytket-qir/pull/147/commits/d83c45f7d794586d99beb86556802e1d9d30558c)
[d83c45f](https://github.com/CQCL/pytket-qir/pull/147/commits/d83c45f7d794586d99beb86556802e1d9d30558c)

cqc-melf: I have removed the check option from the api.
The not time extensive checks are still in and run for all input.
The other checks are removed.
See @cqc-melf
[remove check option from api](https://github.com/CQCL/pytket-qir/pull/147/commits/280d89c8960d1459356607c04ca7c50fe133e4b5)
[280d89c](https://github.com/CQCL/pytket-qir/pull/147/commits/280d89c8960d1459356607c04ca7c50fe133e4b5)

I am still not sure if this is the best option, we could have long running conversion fail late which could be prevented with the check early on.

tests/api_test.py
```diff
@@ -71,6 +71,18 @@ def test_pytket_api_creg() -> None:
...
 
+def test_pytket_api_creg_ii() -> None:
+    circ = Circuit(3)
+
+    circ.add_bit(Bit("c2", 3))
+    circ.add_bit(Bit("c2", 1))
+
+    circ.H(0, condition=Bit("c2", 1))
+
+    with pytest.raises(AssertionError):
```
cqc-alec: This probably shouldn't be an `AssertionError`, it should raise some other exception.

pytket/qir/conversion/api.py
```diff
@@ -46,6 +48,7 @@ def pytket_to_qir(
     wfh: Optional[wasm.WasmFileHandler] = None,
     int_type: int = 64,
     cut_pytket_register: bool = False,
+    check_input: bool = True,
```
cqc-alec: Remove argument here (and from docstring).

cqc-alec (commented):
If time taken to error is a concern, we could move some of the validity checking to earlier in the function (unless that will add a lot of overhead).

pytket/qir/conversion/api.py
cqc-melf: Sorry, I missed that. I will clean this up at the end.

cqc-melf:
> If time taken to error is a concern, we could move some of the validity checking to earlier in the function (unless that will add a lot of overhead).

It was brought up on Tuesday that this is a concern, that is why I had moved the test to the beginning and made it optional.
I made it optional because it takes quiet some time compared to other operations, for the circuit I looked at it generated additional ~5% / ~1min runtime. I am not sure how much this additional runtime is a concern. If this is not a problem I would just put it back in, if this is a problem, I am happy to update register calls and conversion to have additional checks for the incomplete registers to make sure they fails with a useful error message every time.

My personal opinion is that having the check at the beginning with an opt out option would be an acceptable compromise between all this considerations.

tests/api_test.py
cqc-melf: See comment below: https://github.com/CQCL/pytket-qir/pull/147#issuecomment-2196798031

cqc-alec:
> > If time taken to error is a concern, we could move some of the validity checking to earlier in the function (unless that will add a lot of overhead).
> 
> It was brought up on Tuesday that this is a concern, that is why I had moved the test to the beginning and made it optional. I made it optional because it takes quiet some time compared to other operations, for the circuit I looked at it generated additional ~5% / ~1min runtime. I am not sure how much this additional runtime is a concern. If this is not a problem I would just put it back in, if this is a problem, I am happy to update register calls and conversion to have additional checks for the incomplete registers to make sure they fails with a useful error message every time.
> 
> My personal opinion is that having the check at the beginning with an opt out option would be an acceptable compromise between all this considerations.

Another option would be to have another function like `validate_circuit()` that just performs these checks and which the user can choose to call before conversion.

I would be surprised if the time taken to report errors for huge invalid circuits were a major issue. But happy to go with whichever option you decide on.

pytket/qir/conversion/api.py
cqc-melf: Removed in @cqc-melf
[add check function and clean up](https://github.com/CQCL/pytket-qir/pull/147/commits/6a78ce73daff7ad84672e01b40030fd2c83a46cd)
[6a78ce7](https://github.com/CQCL/pytket-qir/pull/147/commits/6a78ce73daff7ad84672e01b40030fd2c83a46cd)

tests/api_test.py
cqc-melf: I have added the additional error messages and updated the testcases.
See details in @cqc-melf
[add check function and clean up](https://github.com/CQCL/pytket-qir/pull/147/commits/6a78ce73daff7ad84672e01b40030fd2c83a46cd)
[6a78ce7](https://github.com/CQCL/pytket-qir/pull/147/commits/6a78ce73daff7ad84672e01b40030fd2c83a46cd)

cqc-melf:
> > > If time taken to error is a concern, we could move some of the validity checking to earlier in the function (unless that will add a lot of overhead).
> > 
> > 
> > It was brought up on Tuesday that this is a concern, that is why I had moved the test to the beginning and made it optional. I made it optional because it takes quiet some time compared to other operations, for the circuit I looked at it generated additional ~5% / ~1min runtime. I am not sure how much this additional runtime is a concern. If this is not a problem I would just put it back in, if this is a problem, I am happy to update register calls and conversion to have additional checks for the incomplete registers to make sure they fails with a useful error message every time.
> > My personal opinion is that having the check at the beginning with an opt out option would be an acceptable compromise between all this considerations.
> 
> Another option would be to have another function like `validate_circuit()` that just performs these checks and which the user can choose to call before conversion.
> 
> I would be surprised if the time taken to report errors for huge invalid circuits were a major issue. But happy to go with whichever option you decide on.

I think having an additional function is also a good idea. I have added that, should be in 
[6a78ce7](https://github.com/CQCL/pytket-qir/pull/147/commits/6a78ce73daff7ad84672e01b40030fd2c83a46cd)

I think this should address all the comment, if I have messed anything, please let me know.

pytket/qir/conversion/api.py
```diff
@@ -124,6 +107,49 @@ def pytket_to_qir(
...
+    problems.
+    :param circuit: given circuit
+    :param int_type: size of each integer, allowed value 32 and 64
+    """
+    _check_circuit_before_conversion(circuit, int_type)
+
+    set_circ_register = set([creg.name for creg in circuit.c_registers])
+    for b in set([b.reg_name for b in circuit.bits]):
+        if b not in set_circ_register:
+            raise ValueError(f"Used register {b} in not a valid register")
```
cqc-alec, line 145: Looking at this section of code, I do find it surprising that it adds a significant amount to the runtime.

pytket/qir/conversion/api.py
```diff
@@ -124,6 +107,49 @@ def pytket_to_qir(
...
+def check_circuit(
+    circuit: Circuit,
+    int_type: int = 64,
+) -> None:
+    """runs the checks for the conversion. Running this check
+     is recomended for big circuits that take a long time to be converted.
+    This functions will throw error with suggestion on how to resolve the
+    problems.
+    :param circuit: given circuit
+    :param int_type: size of each integer, allowed value 32 and 64
```
```suggestion
    """Checks the validity of the circuit.
    
    Running this check before conversion is recommended for big circuits that
    take a long time to be converted.

    :param circuit: given circuit
    :param int_type: integer bit width (32 or 64)
    :raises ValueError: with a suggestion on how to resolve the problems
```

pytket/qir/conversion/api.py
cqc-melf: I think this is because of the circuit structure I am testing, it has a lot of classic registers. Happy to do another measurement of the runtime if you think that would be good.

cqc-alec: I'd be interested to see the numbers. If it's almost always small compared with the time taken for `pytket_to_qir()` then I'd be inclined to put all the checks into that function.

cqc-melf: The difference was only showing up for the big circuit, I have not checked the runtime for the small circuits, but I would expect that the difference is below 0.001 for them.
I will run the profiling for both cases again to make sure there was nothing interfering with the measurements last time.
What do you mean with `put all the checks into that function`?

cqc-alec: > What do you mean with `put all the checks into that function`?

I mean, have `pytket_to_qir()` do all sanity checks early, by calling the `check_circuit()` function. (We wouldn't then need the underscored function which it calls currently.)

cqc-melf: I have run a detailed profile for the check function, the runtime for the additional register check is ~1s now. I am not sure what changed to the other run I did.

I have clean up the check function, it is now always fully run in the conversion.

See [66f7f02](https://github.com/CQCL/pytket-qir/pull/147/commits/66f7f02ab4f176ac4e74a35eda71ba7598d87bb4)

cqc-alec (approved):

PR #163, opened 2024-08-13
add phi nodes

cqc-melf:
# Description

This PR contains the second part of add phi nodes.

There are still 4 functions that needs to be removed to finish this.

# Related issues

# Checklist

- [x] I have made corresponding changes to the public API documentation.
- [x] I have updated the changelog with any user-facing changes.

### DIFF
--- a/pytket/qir/conversion/api.py
+++ b/pytket/qir/conversion/api.py
@@ -60,7 +60,7 @@ def pytket_to_qir(
     :param int_type: size of each integer, allowed value 32 and 64
     :param cut_pytket_register: breaks up the internal scratch bit registers
       into smaller registers, default value false
-    :param profile: generates qir corresponding to the standart profile
+    :param profile: generates QIR corresponding to the standard profile
     """
 
     if cut_pytket_register:
--- a/pytket/qir/conversion/api.py
+++ b/pytket/qir/conversion/api.py
@@ -29,6 +29,7 @@ from pytket.passes import (
 
 from .conversion import QirGenerator
 from .module import tketqirModule
+from .pconversion import PQirGenerator
 
 
 class QIRFormat(Enum):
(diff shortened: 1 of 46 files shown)


cqc-melf:
@cqc-alec Feel free to take a look at this now, I will add the last part in the next days.
I will replace all the four functions for setting and getting bits from the i64 in the last step.

pytket/qir/conversion/api.py
```diff
@@ -58,6 +60,7 @@ def pytket_to_qir(
     :param int_type: size of each integer, allowed value 32 and 64
     :param cut_pytket_register: breaks up the internal scratch bit registers
       into smaller registers, default value false
+    :param profile: generates qir corresponding to the standart profile
```
```suggestion
    :param profile: generates QIR corresponding to the standard profile
```
What is the "standard profile"? Is there a reference for this? And what profile is generated if the flag is not set? I think some more explanation is needed.

pytket/qir/conversion/pconversion.py
cqc-alec, line 1: There is a great deal of code duplication between this file and `conversion.py` which I'd prefer to avoid. Is the plan later to merge the two?

pytket/qir/conversion/api.py
cqc-melf: I will add some more details.

pytket/qir/conversion/pconversion.py
cqc-melf: The idea is to keep both until the profile way fully works and supports all the features we want and then drop the other. Not sure how soon that will be. I would expect at least one release with both. I thought it is the best way to have two separate files to make sure we can do a clean cut if we want to.
If that is to much code duplication now, I am happy to merge both files.

cqc-alec: OK, if we're planning to remove the old one then I think it's fine to keep both for now.

pytket/qir/conversion/api.py
cqc-melf: The new adaptive profile is not yet online, I have added two links in [904c87f](https://github.com/CQCL/pytket-qir/pull/163/commits/904c87faa156d11d43d83eb611123798d1ece4a2)

cqc-melf:
I would like to merge this PR without adding the `zext` and `trunc` operations to have separate release candidates and make sure we can test this individually.

cqc-alec (approved):

PR #177, opened 2024-10-21
add base profile

cqc-melf:
# Description

Add support for the base profile

# Related issues

# Checklist

### DIFF
--- a/pytket/qir/conversion/baseprofileqirgenerator.py
+++ b/pytket/qir/conversion/baseprofileqirgenerator.py
@@ -75,7 +75,7 @@ class BaseProfileQirGenerator(AbstractQirGenerator):
         )
 
     def _get_bit_from_creg(self, creg: str, index: int) -> Value:
-        raise ValueError("Classical REG Ops Not available in base profile")
+        raise ValueError("Classical register operations not available in base profile")
 
     def _set_bit_in_creg(self, creg: str, index: int, ssa_bit: Value) -> None:
         raise ValueError("Classical REG Ops Not available in base profile")
@@ -96,7 +96,7 @@ class BaseProfileQirGenerator(AbstractQirGenerator):
         raise ValueError("Classical REG Ops Not available in base profile")
 
     def conv_conditional(self, command: Command, op: Conditional) -> None:
-        raise ValueError("Conditional NOT available in base profile")
+        raise ValueError("Conditional not available in base profile")
 
     def conv_measure(self, bits: list[Bit], qubits: list[Qubit]) -> None:
         assert len(bits) == 1
--- a/pytket/qir/conversion/baseprofileqirgenerator.py
+++ b/pytket/qir/conversion/baseprofileqirgenerator.py
@@ -1,0 +1,126 @@
+# Copyright 2019-2024 Quantinuum
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Optional
+
+import pyqir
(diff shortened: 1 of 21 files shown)


tests/qir/test_pytket_qir_20-QIRProfile.BASE.ll
```diff
@@ -0,0 +1,71 @@
+; ModuleID = 'test_pytket_qir_20'
+source_filename = "test_pytket_qir_20"
+
+%Qubit = type opaque
+%Result = type opaque
+
+@0 = internal constant [5 x i8] c"c[0]\00"
```
cqc-melf, line 7: We need to record the results individually, there might be device specific options to do this is a better way. I think this is the best way to do it compatible to the base profile.

pytket/qir/conversion/baseprofileqirgenerator.py
```diff
@@ -0,0 +1,126 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Optional
+
+import pyqir
+from pyqir import Value
+
+from pytket import Bit, Circuit, Qubit, wasm  # type: ignore
```
cqc-alec: Prefer `from pytket.wasm import WasmFileHandler` and use that directly; then we may not need this "type: ignore"?

pytket/qir/conversion/baseprofileqirgenerator.py
```diff
@@ -0,0 +1,126 @@
...
+        raise ValueError("Classical REG Ops Not available in base profile")
+
+    def set_ssa_vars(self, reg_name: str, ssa_i64: Value, trunc: bool) -> None:
+        raise ValueError("Classical REG Ops Not available in base profile")
+
+    def _reg2ssa_var(self, bit_reg: BitRegister) -> Value:
+        raise ValueError("Classical REG Ops Not available in base profile")
+
+    def conv_conditional(self, command: Command, op: Conditional) -> None:
+        raise ValueError("Conditional NOT available in base profile")
```
cqc-alec:
```suggestion
        raise ValueError("Conditional not available in base profile")
```

pytket/qir/conversion/baseprofileqirgenerator.py
```diff
@@ -0,0 +1,126 @@
...
+                pyqir.Type.void(self.module.module.context),
+                [
+                    pyqir.result_type(self.module.module.context),
+                    pyqir.PointerType(pyqir.IntType(self.module.module.context, 8)),
+                ],
+            ),
+        )
+
+    def _get_bit_from_creg(self, creg: str, index: int) -> Value:
+        raise ValueError("Classical REG Ops Not available in base profile")
```
cqc-alec:
```suggestion
        raise ValueError("Classical register operations not available in base profile")
```
(similarly below)

pytket/qir/conversion/baseprofileqirgenerator.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def _reg2ssa_var(self, bit_reg: BitRegister) -> Value:
+        raise ValueError("Classical REG Ops Not available in base profile")
+
+    def conv_conditional(self, command: Command, op: Conditional) -> None:
+        raise ValueError("Conditional NOT available in base profile")
+
+    def conv_measure(self, bits: list[Bit], qubits: list[Qubit]) -> None:
+
+        qubit_index = qubits[0].index[0]
```
cqc-alec, line 108: Should probably (a) check that `bits` and `qubits` are the same length and (b) handle the case where that length is 0.

cqc-melf: They should always have exactly one entry, I have added an assert. See [40b2085](https://github.com/CQCL/pytket-qir/pull/177/commits/40b2085eabba5552288d5969a799063d171e15eb)

pytket/qir/conversion/baseprofileqirgenerator.py
cqc-melf: I have changed that, see [195660a](https://github.com/CQCL/pytket-qir/pull/177/commits/195660a65126a4787976751d51e98bc49fe8c481)
The changes needed to make mypy happy ended up a bit bigger.

cqc-alec (approved):

PR #185, opened 2024-10-29
Add support for `ClExprOp` conversion

### LINKED ISSUE #179: Add support for converting `ClExprOp`

cqc-alec:
Closes #179 .


tests/qasm_test.py
```diff
@@ -15,14 +15,37 @@
...
 
 
+def test_pytket_qir_qasm_classical_0() -> None:
+    with open("qasm/test0.qasm") as my_file:
+        circ = circuit_from_qasm_str(my_file.read(), use_clexpr=True)
+
+    run_qir_gen_and_check(
+        circ,
+        "test_pytket_qir_qasm_classical_0",
+        profile=QIRProfile.PYTKET,
```
cqc-melf: Is there a reason why you are only checking the pytket profile in this tests?

pytket/qir/conversion/qirgenerator.py
```diff
@@ -113,6 +119,60 @@
     BitZero: 0,
 }
 
+_TK_CLEXPR_OP_TO_PYQIR: dict = {
+    ClOp.BitAnd: lambda b: b.and_,
+    ClOp.BitOr: lambda b: b.or_,
+    ClOp.BitXor: lambda b: b.xor,
+    ClOp.BitEq: lambda b: partial(b.icmp, IntPredicate.EQ),
+    ClOp.BitNeq: lambda b: partial(b.icmp, IntPredicate.NE),
+    # ClOp.BitNot
```
cqc-melf, line 128: Are all of this ops implemented later on differently? Or are some of them not supported?

pytket/qir/conversion/qirgenerator.py
```diff
@@ -731,6 +791,153 @@ def conv_classicalexpbox(self, op: ClassicalExpBox, args: list) -> None:
...
+            match expr_op:
+                case ClOp.BitAnd | ClOp.BitOr | ClOp.BitXor | ClOp.BitEq | ClOp.BitNeq:
+                    return True, _TK_CLEXPR_OP_TO_PYQIR[expr_op](self.module.builder)(
+                        *ssa_args
+                    )
+                case ClOp.BitNot:
+                    # Implemented as x --> 1 - x
+                    assert len(ssa_args) == 1
+                    return True, self.module.builder.sub(
+                        pyqir.const(self.qir_bool_type, 1), ssa_args[0]
```
cqc-melf, line 838: I think we should put in a request for pyqir to support not operation...
Happy to do that.

pytket/qir/conversion/qirgenerator.py
```diff
@@ -731,6 +791,153 @@ def conv_classicalexpbox(self, op: ClassicalExpBox, args: list) -> None:
...
+                        self.qir_int_type, (1 << self.int_size) - 1
+                    )
+                case ClOp.RegNeg:
+                    # Implemented as x --> 0 - x
+                    assert len(ssa_args) == 1
+                    return False, self.module.builder.sub(
+                        pyqir.const(self.qir_int_type, 0), ssa_args[0]
+                    )
+                case ClOp.RegDiv | ClOp.RegPow:
+                    raise ValueError(f"Classical operation {expr_op} not supported")
```
cqc-melf, line 890: Do we want to link the issues here?

setup.py
```diff
@@ -45,7 +45,7 @@
     packages=find_namespace_packages(include=["pytket.*"]),
     include_package_data=True,
     install_requires=[
-        "pytket >= 1.33.0",
+        "pytket >= 1.34.0",
```
cqc-melf, line 48: Do you want to put this in the changelog?
Maybe adding support for the new clexp as well?

tests/qasm_test.py
cqc-alec: When I use the `ADAPTIVE` profile most of the classical logic is optimized away so it's harder to check the result. (I'm not sure why this isn't the case with the `PYTKET` profile.) But I can add tests using the `ADAPTIVE` profile as well.

pytket/qir/conversion/qirgenerator.py
cqc-alec: Some are implemented differently; a couple of them are unsupported (these are `RegDiv` and `RegPow` which are also unsupported for `logic_exp` conversion; there is an issue for this).

pytket/qir/conversion/qirgenerator.py
cqc-alec: I'll insert a comment with the issue URL.

tests/qasm_test.py
cqc-melf: Pyqir does not know how to interpret the classical functions we use for the registers, so it can't optimise them. I think it is good to have the optimised result, too.

cqc-melf (approved):

PR #189, opened 2024-12-04
add azure option in result recording

cqc-melf:
# Description

Please summarise the changes.

# Related issues

# Checklist

### DIFF
--- a/pytket/qir/conversion/api.py
+++ b/pytket/qir/conversion/api.py
@@ -26,6 +26,8 @@ from pytket.passes import (
     scratch_reg_resize_pass,
 )
 
+from .azurebaseprofileqirgenerator import AzureBaseProfileQirGenerator
+from .azureprofileqirgenerator import AzureAdaptiveProfileQirGenerator
 from .baseprofileqirgenerator import BaseProfileQirGenerator
 from .module import tketqirModule
 from .profileqirgenerator import AdaptiveProfileQirGenerator
@@ -48,9 +50,11 @@ class QIRProfile(Enum):
     """Profile for the QIR generation"""
 
     BASE = 0
-    ADAPTIVE = 1
-    ADAPTIVE_CREGSIZE = 2
-    PYTKET = 3
+    AZUREBASE = 1
+    ADAPTIVE = 2
+    AZUREADAPTIVE = 3
+    ADAPTIVE_CREGSIZE = 4
+    PYTKET = 5
 
 
 def pytket_to_qir(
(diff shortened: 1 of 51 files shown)


pytket/qir/conversion/api.py
```diff
@@ -122,12 +133,42 @@ def pytket_to_qir(
...
+        )
     else:
         raise NotImplementedError("unexpected profile")
 
     populated_module = qir_generator.circuit_to_module(qir_generator.circuit, True)
 
-    if qir_generator.has_wasm:
+    if profile == QIRProfile.AZUREADAPTIVE:
+
+        assert not qir_generator.has_wasm
```
cqc-melf, line 151: My understanding is that we don't want to do wasm for azure.

cqc-alec: I think that's a fair assumption for now at least.

pytket/qir/conversion/api.py
```diff
@@ -122,12 +133,42 @@ def pytket_to_qir(
...
+        sar_azure_dict: dict[str, str] = qir_generator.get_azure_sar()
+
+        initial_result = str(populated_module.module.ir())
+
+        for az in sar_azure_dict:
+            initial_result = initial_result.replace(az, sar_azure_dict[az])
+
+        result = initial_result
+
+        bitcode = pyqir.Module.from_ir(pyqir.Context(), result).bitcode
```
cqc-alec, line 162: May as well only compute this if `qir_format == QIRFormat.BINARY`.

pytket/qir/conversion/azurebaseprofileqirgenerator.py
```diff
@@ -0,0 +1,89 @@
...
+        )
+
+    def conv_measure(self, bits: list[Bit], qubits: list[Qubit]) -> None:
+        assert len(bits) == 1
+        assert len(qubits) == 1
+
+    def record_output(self) -> None:
+
+        # this will measure all qubits at the end of the circuit
+        # the result of the measurement will be added to an array and recorded togeather
```
cqc-alec:
```suggestion
        # the result of the measurement will be added to an array and recorded together
```

pytket/qir/conversion/api.py
cqc-melf: I think it would be good to keep that. That way we make sure that our search and replace result is still parable for pyqir and we don't submit anything invalid.

cqc-alec (approved):

PR #198, opened 2025-02-25
Fix handling of register predicate ops in classical expressions

### LINKED ISSUE #197: Cannot convert `ClOp.RegEq` and similar to QIR
```python
from pytket.circuit import Circuit
from pytket.circuit.clexpr import wired_clexpr_from_logic_exp
from pytket.circuit.logic_exp import reg_eq
from pytket.qir.conversion.api import QIRFormat, QIRProfile, pytket_to_qir

circ = Circuit(2)
a = circ.add_c_register("a", 2)
b = circ.add_c_register("b", 2)
c = circ.add_c_register("c", 2)

wexpr, args = wired_clexpr_from_logic_exp(reg_eq(a, b), c)
circ.add_clexpr(wexpr, args)

# This works:
# circ.add_classicalexpbox_register(reg_eq(a, b), c)


gen_qir_ll = pytket_to_qir(
    circ,
    qir_format=QIRFormat.STRING,
    profile=QIRProfile.ADAPTIVE,
    cut_pytket_register=True,
)
```
The above code raises an exception:

```
ValueError: Call parameter type does not match function signature!
i1 true
 i64  call void @__quantum__rt__int_record_output(i1 true, i8* getelementptr inbounds ([2 x i8], [2 x i8]* @2, i32 0, i32 0))
```

With the (deprecated) `add_classicalexpbox_register` form it works.

It looks like the return type of the comparison operation is `i64` when it is expected to be `i1`.

cqc-alec:
Fixes #197 .

### DIFF
--- a/tests/conversion_test.py
+++ b/tests/conversion_test.py
@@ -16,6 +16,7 @@ import pytest
 from utilities import run_qir_gen_and_check  # type: ignore
 
 from pytket.circuit import Bit, BitRegister, Circuit, Qubit, if_not_bit
+from pytket.circuit.clexpr import wired_clexpr_from_logic_exp
 from pytket.circuit.logic_exp import (
     reg_eq,
     reg_geq,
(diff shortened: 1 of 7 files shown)


tests/conversion_test.py
```diff
@@ -546,6 +547,32 @@ def test_pytket_qir_20(profile: QIRProfile) -> None:
...
+    ],
+)
+def test_pytket_qir_21(profile: QIRProfile) -> None:
+    # test classical expression handling with register predicates
+    circ = Circuit(2)
+    a = circ.add_c_register("a", 3)
+    b = circ.add_c_register("b", 3)
+    c = circ.add_c_register("c", 1)
+
+    circ.add_clexpr(*wired_clexpr_from_logic_exp(reg_eq(a, b), c))  # type: ignore
```
cqc-melf, line 567: Why do we have to type ignore this? Is this not typed in pytket?

tests/conversion_test.py
```diff
@@ -546,6 +547,32 @@ def test_pytket_qir_20(profile: QIRProfile) -> None:
...
+@pytest.mark.parametrize(
+    "profile",
+    [
+        QIRProfile.ADAPTIVE,
+        QIRProfile.AZUREADAPTIVE,
+        QIRProfile.PYTKET,
+        QIRProfile.ADAPTIVE_CREGSIZE,
+    ],
+)
+def test_pytket_qir_21(profile: QIRProfile) -> None:
```
cqc-melf, line 559: This testcase is not really ideal because most of the classical operation disappear after the classical optimisation. Could you maybe add one initial quantum measurement to one of the registers so we see the classical ops in the QIR?

tests/conversion_test.py
cqc-alec: For the same reason that we have to [here](https://github.com/CQCL/pytket-qir/blob/6df5e1b633e60e23d660b29defe6fe6520681181/tests/conversion_test.py#L193) (this test was based on that one). I think it may be to do with implicitly casting a `BitRegister` to a `list`.

tests/conversion_test.py
cqc-alec: Yes, will do.

cqc-alec: Done. Are the QIR files as you would expect?

tests/conversion_test.py
cqc-melf: Is there something we can do about this, or do we just accept that?

tests/conversion_test.py
cqc-melf: Looks good!

cqc-melf (approved):

PR #205, opened 2025-03-12
Fix formatting in docstring.

### LINKED ISSUE #199: Improve formatting in documentation of `pytket_to_qir`
The formatting of the description of the `profile` parameter should be improved, e.g. with bullets:

Image

cqc-alec:
Fixes #199 .

Now looks like this:

image

### DIFF
--- a/pytket/qir/conversion/api.py
+++ b/pytket/qir/conversion/api.py
@@ -87,6 +87,10 @@ def pytket_to_qir(
         https://github.com/qir-alliance/qir-spec/tree/main/specification/under_development/profiles/Adaptive_Profile.md
       - Use ``QIRProfile.PYTKET`` for QIR with additional functions for classical
         registers.
+      - Use ``QIRProfile.AZUREBASE`` for the base profile with metadata for Azure
+        devices and ``array_record`` for output recording.
+      - Use ``QIRProfile.AZUREADAPTIVE`` for the adaptive profile with metadata for
+        Azure devices and bitwise recording of results via ``array_record``.
     """
 
     if cut_pytket_register:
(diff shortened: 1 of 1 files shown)


cqc-melf:
I think we decided to not have AZUREBASE and AZUREADAPTIVE in the list because they were expected to change. As are using them both now in pytket-azure we could add them in the list here?

cqc-melf (commented):
Looks good. 

I think we decided to not have AZUREBASE and AZUREADAPTIVE in the list because they were expected to change. As are using them both now in pytket-azure we could add them in the list here?

cqc-alec:
> Looks good.
> 
> I think we decided to not have AZUREBASE and AZUREADAPTIVE in the list because they were expected to change. As are using them both now in pytket-azure we could add them in the list here?

Do you have some text and/or links for those?

pytket/qir/conversion/api.py
```diff
@@ -76,16 +76,17 @@ def pytket_to_qir(
...
-        truncation operation to assure that integers matching the classical
-        registers have no unexpected set bits, see:
+      - Use ``QIRProfile.ADAPTIVE_CREGSIZE`` for the adaptive profile with additional
+        truncation operations to assure that integers matching the classical
+        registers have no unexpected set bits. See:
         https://github.com/qir-alliance/qir-spec/tree/main/specification/under_development/profiles/Adaptive_Profile.md
-        Use QIRProfile.PYTKET for QIR with additonal function for classical registers.
-
+      - Use ``QIRProfile.PYTKET`` for QIR with additional functions for classical
+        registers.
```
cqc-melf, line 89:
```suggestion
        registers.
      - Use ``QIRProfile.AZUREBASE`` for the base profile with meta data for the azure devices and array_record for the output recording.
      - Use ``QIRProfile.AZUREADAPTIVE`` for the adaptive profile with meta for the azure devices and bitwise recording of the results via array_record.

cqc-melf:
> > Looks good.
> > I think we decided to not have AZUREBASE and AZUREADAPTIVE in the list because they were expected to change. As are using them both now in pytket-azure we could add them in the list here?
> 
> Do you have some text and/or links for those?

I have added a suggestion, I think that should do it. Happy to expand it.

cqc-melf (approved):

PR #215, opened 2025-05-08
add qasm tests

cqc-melf:
# Description

Please summarise the changes.

# Related issues

# Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ setup(
     packages=find_namespace_packages(include=["pytket.*"]),
     include_package_data=True,
     install_requires=[
-        "pytket >= 2.2.0",
+        "pytket >= 2.4.0rc0",
         "pyqir ~= 0.10.9",
     ],
     classifiers=[
--- a/tests/qir/test_bitreg_as_bit_8-QIRProfile.PYTKET.ll
+++ b/tests/qir/test_bitreg_as_bit_8-QIRProfile.PYTKET.ll
@@ -1,0 +1,48 @@
+; ModuleID = 'test_bitreg_as_bit_8'
+source_filename = "test_bitreg_as_bit_8"
+
+%Result = type opaque
+%Qubit = type opaque
+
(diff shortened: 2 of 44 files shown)


tests/qir/test_bitreg_as_bit_8-QIRProfile.PYTKET.ll
```diff
@@ -0,0 +1,48 @@
+; ModuleID = 'test_bitreg_as_bit_8'
+source_filename = "test_bitreg_as_bit_8"
+
+%Result = type opaque
+%Qubit = type opaque
+
+@0 = internal constant [2 x i8] c"a\00"
```
cqc-alec, line 7: Is this expected?

setup.py
```diff
@@ -45,7 +45,7 @@
     packages=find_namespace_packages(include=["pytket.*"]),
     include_package_data=True,
     install_requires=[
-        "pytket >= 2.2.0",
+        "pytket >= 2.4.0rc0",
```
cqc-alec, line 48: I'm not sure we should merge this until we have a proper release we can depend on.

cqc-melf: I would suggest we merge this now, but wait with a release?

tests/qir/test_bitreg_as_bit_8-QIRProfile.PYTKET.ll
cqc-melf: Yes, is it only this line you are worried about? Which point are you worried here?

cqc-alec: I was misreading it, thought `a` was being created as a register of size 2, but this is just the name, OK.

setup.py
cqc-alec: OK, as long as we remember to update this dependency when we do a release.

cqc-alec (approved):

PR #219, opened 2025-05-28
Add `ClassicalRegisterWidthError` to API

cqc-alec:
In order to address https://github.com/CQCL/pytket-quantinuum/issues/589 .

### DIFF
--- a/pytket/qir/conversion/api.py
+++ b/pytket/qir/conversion/api.py
@@ -57,6 +57,15 @@ class QIRProfile(Enum):
     PYTKET = 5
 
 
+class ClassicalRegisterWidthError(Exception):
+    """Error trying to convert a circuit with a classical register exceeding the maximum width"""
+
+    def __init__(self, width: int, max_width: int = 64) -> None:
+        super().__init__(
+            f"Classical register of width {width} exceeds maximum width ({max_width})."
+        )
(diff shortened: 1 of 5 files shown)


pytket/qir/conversion/api.py
```diff
@@ -57,6 +57,15 @@ class QIRProfile(Enum):
     PYTKET = 5
 
 
+class ClassicalRegisterWidthError(Exception):
+    """Error trying to convert a circuit with a classical register exceeding the maximum width"""
+
+    def __init__(self, width: int, max_width: int = 64) -> None:
+        super().__init__(
+            f"Classical register of width {width} exceeds maximum width ({max_width})."
```
cqc-melf, line 67: This part of the error message is lost, should we add that back in?
```
you could try to set cut_pytket_register=True in the conversion

cqc-alec: I think I will add a "hint" parameter to the exception initializer. (Then different hints can be used in different contexts.)

cqc-melf (approved):
If you want to we can release this directly

cqc-alec:
> If you want to we can release this directly

No urgency, but it would allow me to close the pytket-quantinuum issue sooner!

PR #224, opened 2025-07-31
Add gateset check to check circuit

isobelhooper:
# Description

Add an extra test to `check_circuit` to make sure the circuit only contains gates that can be converted to QIR.

It takes a set of supported pytket OpTypes, which defaults to `PYQIR_GATES.base_gateset`.

Also updated the docstring for `check_circuit()` to list both the exceptions that can be raised.

# Related issues

None that I know of.

# Checklist

### DIFF
--- a/pytket/qir/conversion/api.py
+++ b/pytket/qir/conversion/api.py
@@ -17,14 +17,16 @@ from typing import TYPE_CHECKING
 
 import pyqir
 
-from pytket.circuit import Circuit
+from pytket.circuit import Circuit, OpType
 from pytket.passes import (
     scratch_reg_resize_pass,
 )
+from pytket.predicates import GateSetPredicate
 
 from .azurebaseprofileqirgenerator import AzureBaseProfileQirGenerator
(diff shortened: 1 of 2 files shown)


pytket/qir/conversion/api.py
```diff
@@ -247,3 +252,7 @@ def check_circuit(
     for b in {b.reg_name for b in circuit.bits}:
         if b not in set_circ_register:
             raise ValueError(f"Used register {b} in not a valid register")
+
+    gate_set_predicate = GateSetPredicate(gate_set)
+    if not gate_set_predicate.verify(circuit):
+        raise ValueError(f"Circuit contains gates that can't be converted to QIR. Supported gates: {gate_set.base_gateset}")
```
cqc-melf:
```suggestion
        raise ValueError(f"Circuit contains gates that can't be converted to QIR. Supported gates: {gate_set}")

cqc-melf:
I think wasm is not in the gateset, could you inject that later on into the predicate?

PR #250, opened 2026-01-09
chore: upgrade pyqir to ~0.11, update dev deps

qartik:


pyproject.toml
```diff
@@ -18,6 +18,7 @@ classifiers = [
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
```
cqc-melf: I would suggest that we don't add this for now as it is not yet supported on the CI for testing? (See the commit I tried)

docs/changelog.rst
```diff
@@ -3,6 +3,11 @@
 Changelog
 ~~~~~~~~~
 
+0.27.0
```
cqc-melf, line 6: I have added an entry to the changelog

cqc-melf (changes requested):

pyproject.toml
qartik:
```suggestion
```

cqc-alec (approved):

PR #253, opened 2026-02-27
update version 1.0.0

cqc-melf:
# Description

This will be the last release for QIR 1.0

# Checklist

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -3,9 +3,10 @@
 Changelog
 ~~~~~~~~~
 
-0.27.1
-------
+1.0.0 (February 2026)
+---------------------
 
+* Updated pytket version requirement to 2.13.
 * Update pyqir minimium version requirement to 0.11.1.
 
 0.26.0 (January 2026)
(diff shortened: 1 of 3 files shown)


cqc-alec:
What are the plans for versioning of future releases?

cqc-melf:
> What are the plans for versioning of future releases?

1.x -> for QIR 1.0
2.x -> for QIR 2.0

cqc-alec:
That means that if we follow semver and want to make a breaking change to either 1.x or 2.x we'll need to change the major version, which could be quite confusing. We could alternatively consider a parameter that selects which version of QIR to output.

cqc-alec:
Or we could have two separate packages.

cqc-alec:
I think the approach here would be fine if we are basically deprecating QIR 1.0 generation now.

cqc-melf:
The plan is to only make 1.x releases in case of very serious issues. There are no new features planed for 1.x

We can't do a parameter to select the QIR version, because pyqir > 0.11 does not support the generation of QIR 1.0 anymore.

cqc-alec (approved):

